PR #11, opened 2015-01-30
Add PHP composer comparison scripts.

jvkersch:


simplesat/scripts/yaml_to_repository.py
```diff
@@ -0,0 +1,43 @@
...
+        package = parser.parse_to_package(package_str, "2.7")
+        repository.add_package(package)
+
+    return repository
+
+
+def main():
+    p = argparse.ArgumentParser()
+    p.add_argument("scenario", help="Yaml scenario file")
+    ns = p.parse_args(sys.argv[1:])
```
jvkersch: Good call, I don't really remember why. I'll fix it.

PR #35, opened 2015-09-28
Add installed packages rules

cournape:
We now add rules for installed packages, of the form ` A | B | C | ...` where `A`, `B`, `C`, etc.. are the available versions of installed package `p`.


cournape:
@jvkersch if you try to solve the added scenarios, you will see that 2 of them crash the solver (both are unsatisfiable, and are cases where we will want to find the minimal unsatisfiable solution(s))

jvkersch:
LGTM

PR #36, opened 2015-09-30
Cleanup internal org

cournape:
Reorganize and rationalize a bit the package API:

* purely SAT logic is now in `simplesat.sat`
* rest of the code stays in `simplesat` itself
* removed old solver code, and move 0install-based algorithm in old directory

@jvkersch


simplesat/add_clause_script.py
```diff
@@ -1,5 +0,0 @@
-from simplesat.api import Solver, Clause
```
jvkersch: I don't even remember where this came from ;-)

jvkersch:
LGTM.

Travis builds weren't running anymore since the repo was open-sourced. I enabled builds on the open-source Travis infrastructure and kicked off a test run.

jvkersch:
... which fails because of the dependency on enstaller.

simplesat/add_clause_script.py
cournape: Let's pretend it never happened :)

cournape:
I put some commits I should not have, recreating a new PR.

PR #49, opened 2015-10-12
MAINT: move _composer_utils back to simplesat.

cournape:
It will be easier to track composer in one package than two.

It also updates the code to work with more recent versions of composer.


cournape:
@jvkersch

jvkersch:
LGTM

PR #50, opened 2015-10-13
MAINT: Add bibliography.

jvkersch:
@cournape 

I would have preferred not reverting to plain ASCII for the names of the Minisat authors, but I don't know if there is a standard encoding we may assume...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,6 +46,10 @@ Bibliography:
 
 On the use of SAT solvers for managing packages:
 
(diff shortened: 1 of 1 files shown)


cournape:
Can you add the reference to 0install website as well ?

jvkersch:
@cournape Done

rkern:
I think UTF-8 is pretty standard for things that will automatically consume the `README`.

PR #53, opened 2015-10-30
BUG: _find_updates should allow lower version to handle downgrades.

cournape:
The `_find_updates` was only considering versions higher than the given package, which prevents `_find_updates` to work with downgrades.

@johntyree @jvkersch

### DIFF
--- a/simplesat/tests/common.py
+++ b/simplesat/tests/common.py
@@ -109,8 +109,8 @@ class Scenario(object):
             if operation["kind"] == "install":
                 operations.append(InstallOperation(operation["package"]))
             elif operation["kind"] == "update":
-                operations.append(UpdateOperation(operation["from"],
-                                                  operation["to"]))
+                operations.append(UpdateOperation(operation["to"],
+                                                  operation["from"]))
             elif operation["kind"] == "remove":
                 operations.append(RemoveOperation(operation["package"]))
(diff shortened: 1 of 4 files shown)


johntyree:
I haven't ventured far outside of `simplesat/sat/`, but this looks like it's probably right :)

Can you point to yaml file I can run to demonstrate the bug?

cournape:
I don't think we have any scenario testing this, but I can add a quick unittest for transaction calculation

johntyree:
`simplesat/tests/numpy_downgrade.yaml` requires this patch to be handled correctly.

johntyree:
Interesting. Looks like Github orders commits by `author date` rather than `commit date`. In `git log`, these commits are ordered `a8ac, 213f, 7050`. I'd call that a bug.

johntyree:
Ah of course!! :bulb:

This is failing because #54 is overzealous. We're pruning away packages that aren't dependencies. 

Instead, we need to make sure we don't prune other versions of relevant packages, otherwise we can never upgrade anything because we can't uninstall the current version.

johntyree:
I've confirmed manually that #60 resolves this, so merging that, then merging master here should be good to go.

johntyree:
#60 was merged so this is ready to go.

PR #54, opened 2015-11-03
Prune unnecessary packages from solution

johntyree:
This PR implements a simple pruning algorithm to avoid installing packages which end up in the solution but which have no packages that depend on them. It recursively gathers up all of the dependencies of each required package and then does an intersection of those packages with the solution (preserving order).

I think it should be possible to avoid this entirely by being smarter with how backtracking is handled and thereby avoiding this altogether, but that sounds hard and this was very easy.

As a bonus, it revealed that the `iris.yaml` scenario incorrectly includes `zlib` in the expected installed packages, even though no one is asking for it. I have updated the scenario so that tox passes.

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -75,8 +75,8 @@ class DependencySolver(object):
         return requirement_ids, list(rules_generator.iter_rules())
 
 
-def _connected_packages(solution, requirement_ids, pool):
-    """ Return packages which are associated with a requirement. """
+def _connected_packages(solution, pkg_ids, pool):
+    """ Return packages in `solution` which are associated with `pkg_ids`. """
 
     # Our strategy is as follows:
     # ... -> pkg.dependencies -> pkg strings -> ids -> _id_to_package -> ...
@@ -102,10 +102,10 @@ def _connected_packages(solution, requirement_ids, pool):
         neighbors = set(package_string_to_id[key] for key in pkg_keys)
         return neighbors
 
-    # A requirement can root its own independent graph, so we must start at
+    # Each package can root its own independent graph, so we must start at
     # each one individually
     connected = set()
-    for pkg_id in requirement_ids:
+    for pkg_id in pkg_ids:
         # We pass in `connected` to avoid re-walking a graph we've seen before
         connected.update(_connected_nodes(pkg_id, neighborfunc, connected))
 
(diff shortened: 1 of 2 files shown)


simplesat/dependency_solver.py
```diff
@@ -65,12 +71,62 @@ def _create_rules(self, request):
...
+    for pkg_id in sol_set:
+        pkg = pool._id_to_package[pkg_id]
+        pkg_key = pkg.name
+        package_string_to_id[pkg_key] = pkg_id
+
+    def neighborfunc(pkg_id):
+        """ Given a pkg id, return the pkg ids of the dependencies that
+        appeared in our solution. """
+        dep_strings = pool._id_to_package[pkg_id].dependencies
+        pkg_keys = (d.split()[0] for d in dep_strings)
```
cournape: We should not depend on the requirement format being "name version" (which is terrible :) ), but use requirements instead. You can create requirements from that string w/ `requirement = Requirement.from_legacy_requirement_string`, and then take the name with `requirement.name`

simplesat/dependency_solver.py
```diff
@@ -65,12 +71,62 @@ def _create_rules(self, request):
 
         rules_generator = RulesGenerator(pool, request, installed_map)
 
-        return list(rules_generator.iter_rules())
+        return requirement_ids, list(rules_generator.iter_rules())
+
+
+def _connected_packages(solution, requirement_ids, pool):
+    """ Return packages which are associated with a requirement. """
```
cournape: can we call `requirement_ids` `pkg_ids` instead ? IIUC, it does not matter for this function that packages are coming from requirements or not.

cournape:
2 simple style comments, LGTM otherwise

simplesat/dependency_solver.py
johntyree: Definitely agree. At the time I did not know how to extract that in a reasonable way. I meant to add a `FIXME` here and revisit, glad you found it!

johntyree: Why legacy?

johntyree: It's also not clear if you mean `enstaller.solver.Requirement` or `enstaller.newsolver.Requirement`, or indeed when we would ever choose one over the other.

johntyree: Scratch that, obviously you mean `new_solver.Requirement` as it's the only one with that method :)

cournape: @johntyree the dichotomy is gone in the newer enstaller. It is called legacy because the `Requirement` class can handle much richer semantics, and the goal is to support those richer semantics in next egg format (but we need the new solver first :) )

johntyree: Ah I see. OK then.

PR #55, opened 2015-11-03
A faster Installed-First policy

### LINKED ISSUE #42: Crash with simple `>=` request
This scenario is failing:

```
packages:
  - libgfortran 3.0.0-1
  - libgfortran 3.0.0-2
  - MKL 10.2-1
  - MKL 10.2-2
  - MKL 10.3-1
  - numpy 1.5.1-1; depends (MKL == 10.3-1)
  - numpy 1.5.1-2; depends (MKL == 10.3-1)
  - numpy 1.6.0b2-1; depends (MKL == 10.3-1)
  - numpy 1.6.0-1; depends (MKL == 10.3-1)
  - numpy 1.6.0-2; depends (MKL == 10.3-1)
  - numpy 1.6.0-3; depends (MKL == 10.3-1)
  - numpy 1.6.0-4; depends (MKL == 10.3-1)
  - numpy 1.6.0-5; depends (MKL == 10.3-1)
  - numpy 1.6.1-1; depends (MKL == 10.3-1)
  - numpy 1.6.1-2; depends (MKL == 10.3-1)
  - numpy 1.6.1-3; depends (MKL == 10.3-1)
  - numpy 1.6.1-5; depends (MKL == 10.3-1)
  - numpy 1.7.1-1; depends (MKL == 10.3-1)
  - numpy 1.7.1-2; depends (MKL == 10.3-1)
  - numpy 1.7.1-3; depends (MKL == 10.3-1)
  - numpy 1.8.0-1; depends (MKL == 10.3-1)
  - numpy 1.8.0-2; depends (MKL == 10.3-1)
  - numpy 1.8.0-3; depends (MKL == 10.3-1)
  - numpy 1.8.1-1; depends (MKL == 10.3-1)
  - numpy 1.8.1-2; depends (MKL == 10.3-1)
  - numpy 1.8.1-3; depends (libgfortran ~= 3.0.0, MKL == 10.3-1)
  - numpy 1.9.2-1; depends (libgfortran ~= 3.0.0, MKL == 10.3-1)

request:
    - operation: "install"
      requirement: "numpy >= 1.9.2"
````

with the following traceback:

```
  File "scripts/pysolver_policy.py", line 35, in <module>
    main()
  File "scripts/pysolver_policy.py", line 31, in main
    scenario.installed_repository, ns.print_ids)
  File "scripts/pysolver_policy.py", line 16, in solve_and_print
    transaction = solver.solve(request)
  File "/home/davidc/src/enthought/sat-solver/simplesat/pysolver_with_policy.py", line 23, in solve
  File "/home/davidc/src/enthought/sat-solver/simplesat/sat/minisat.py", line 141, in search
    self.clauses,
  File "/home/davidc/src/enthought/sat-solver/simplesat/sat/policy.py", line 64, in get_next_package_id
    "Trying to assign to a variable which is already assigned."
AssertionError: Trying to assign to a variable which is already assigned.
```

### PR DESCRIPTION
**Pull request by johntyree**
This PR introduces a Policy that provides "installed first" behavior to the dependency solver, as well as a few helper classes.

Benchmarks are in issue #52.

It is implemented as a giant priority queue where the priorities split into three bands, in order from highest to lowest:

1. Installed packages
  - Packages which are already installed on the system.
2. Required packages
  - Packages which have been explicitly requested.
  - This is lower priority than the installed packages because we don't want to unnecessarily upgrade things.
3. Everything else.

Within each group, packages are ordered descending by version, such that newer packages are tried first.

I'm currently using the notion that version objects are always comparable to define an ordering over *all* packages and assign priorities based on that, rather than grouping packages by name. In addition to being fast, easy to write, and easy to understand, it leads to a few interesting properties:

### Things it does
-  it suggests packages with large version numbers earlier, i.e. `curl-7.3.0` comes well before `Biggus-0.7.0`.
  - This was very easy to implement, but there's no good reason to do it this way. At best it doesn't hurt us. There are plenty of ways to make this smarter.
- it suggests packages interleaved with each other, i.e. `pep8 1.4.6` then `scipy 0.15.0` then `pep8 0.6.1`,
- high priority packages are suggested repeatedly until all options are exhausted,
- not crash on #42,
- it can totally ignore the `clause` literals and still be quite fast.

To see this play out, there's a `PolicyLogger` class that is currently wrapping `InstalledFirstPolicy` and keeping track of things.

### Things it doesn't do
- Anything smart by examining `clause` instances themselves,
- Anything clever with package metadata (i.e. should we suggest packages with few or many dependencies first?)

It currently suffers from the unrelated backtracking bug that motivated #54, and passes all tox tests with that patch applied.

If this is suitable, I would think that this closes #42 and #52.

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -52,12 +52,7 @@ class DependencySolver(object):
             pool.package_id(package)
             for package in pool.what_provides(requirement)
         ]
-        self._policy.add_packages_by_id(requirement_ids)
-
-        # Add installed packages.
-        self._policy.add_packages_by_id(
-            [pool.package_id(package) for package in installed_repository]
-        )
+        self._policy.add_requirements(requirement_ids)
 
         installed_map = collections.OrderedDict()
         for package in installed_repository:
--- a/simplesat/sat/assignment_set.py
+++ b/simplesat/sat/assignment_set.py
@@ -1,14 +1,12 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-from __future__ import division, print_function
 
-from toolbox import auto_debugger
-
-from collections import OrderedDict
-from heapq import heappush, heappop
-import itertools
 from functools import partial
+from heapq import heappush, heappop
+from itertools import count
+
+import six
 
 
 INF = float('inf')
@@ -24,15 +22,15 @@ class _MISSING(object):
 MISSING = _MISSING()
 
 
-class AssignmentSet(OrderedDict):
+class AssignmentSet(object):
 
     """A collection of literals and their assignments."""
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self):
         self._nassigned = 0
         # Changelog is a dict of id -> (original value, new value)
+        self._data = {}
         self._changelog = {}
-        super(AssignmentSet, self).__init__(*args, **kwargs)
 
     def __setitem__(self, key, value):
         assert key >= 0
@@ -46,20 +44,41 @@ class AssignmentSet(OrderedDict):
             self._nassigned += 1
 
         self._update_changelog(key, value)
-        super(AssignmentSet, self).__setitem__(key, value)
+        self._data[key] = value
 
     def __delitem__(self, key):
         self._update_changelog(key, MISSING)
-        super(AssignmentSet, self).__del__(key)
+        del self._data[key]
 
-    def __str__(self):
-        return object.__str__(self)
+    def __getitem__(self, key):
+        return self._data[key]
+
+    def get(self, key, default=None):
+        return self._data.get(key, default)
+
+    def __len__(self):
+        return len(self._data)
+
+    def __contains__(self, key):
+        return key in self._data
+
+    def items(self):
+        return self._data.items()
+
+    def iteritems(self):
+        return six.iteritems(self._data)
+
+    def keys(self):
+        return self._data.keys()
+
+    def values(self):
+        return self._data.values()
 
     def _update_changelog(self, key, value):
         if key in self._changelog:
             orig = self._changelog[key][0]
-        elif key in self:
-            orig = self[key]
+        elif key in self._data:
+            orig = self._data[key]
         else:
             orig = MISSING
 
@@ -74,9 +93,10 @@ class AssignmentSet(OrderedDict):
         return old
 
     def copy(self):
-        new = super(AssignmentSet, self).copy()
-        new._nassigned = self._nassigned
+        new = AssignmentSet()
+        new._data = self._data.copy()
         new._changelog = self._changelog.copy()
+        new._nassigned = self._nassigned
         return new
 
     @property
@@ -97,7 +117,7 @@ class PriorityQueue(object):
         # mapping of tasks to entries
         self._entry_finder = {}
         # unique id genrator for tie-breaking
-        self._next_id = partial(next, itertools.count())
+        self._next_id = partial(next, count())
 
     def push(self, task, priority=0):
         "Add a new task or update the priority of an existing task"
--- a/simplesat/sat/assignment_set.py
+++ b/simplesat/sat/assignment_set.py
@@ -4,6 +4,8 @@
 
 import six
 
+from collections import OrderedDict
+
 
 class _MISSING(object):
     pass
@@ -17,7 +19,8 @@ class AssignmentSet(object):
     def __init__(self):
         self._nassigned = 0
         # Changelog is a dict of id -> (original value, new value)
-        self._data = {}
+        # FIXME: Verify that we really need ordering here
+        self._data = OrderedDict()
         self._changelog = {}
 
     def __setitem__(self, key, value):
--- a/simplesat/sat/assignment_set.py
+++ b/simplesat/sat/assignment_set.py
@@ -36,7 +36,9 @@ class AssignmentSet(object):
 
     def __delitem__(self, key):
         self._update_changelog(key, MISSING)
-        del self._data[key]
+        prev = self._data.pop(key)
+        if prev is not None:
+            self._nassigned -= 1
 
     def __getitem__(self, key):
         return self._data[key]
--- a/simplesat/sat/assignment_set.py
+++ b/simplesat/sat/assignment_set.py
@@ -16,12 +16,15 @@ class AssignmentSet(object):
 
     """A collection of literals and their assignments."""
 
-    def __init__(self):
+    def __init__(self, assignments=None):
         self._nassigned = 0
         # Changelog is a dict of id -> (original value, new value)
         # FIXME: Verify that we really need ordering here
         self._data = OrderedDict()
-        self._changelog = {}
+        self._orig = {}
+        self._cached_changelog = None
+        for k, v in (assignments or {}).items():
+            self[k] = v
 
     def __setitem__(self, key, value):
         assert key >= 0
@@ -34,11 +37,11 @@ class AssignmentSet(object):
         if value is not None:
             self._nassigned += 1
 
-        self._update_changelog(key, value)
+        self._update_diff(key, value)
         self._data[key] = value
 
     def __delitem__(self, key):
-        self._update_changelog(key, MISSING)
+        self._update_diff(key, MISSING)
         prev = self._data.pop(key)
         if prev is not None:
             self._nassigned -= 1
@@ -67,31 +70,32 @@ class AssignmentSet(object):
     def values(self):
         return list(self._data.values())
 
-    def _update_changelog(self, key, value):
-        if key in self._changelog:
-            orig = self._changelog[key][0]
-        elif key in self._data:
-            orig = self._data[key]
-        else:
-            orig = MISSING
-
-        if orig == value:
-            self._changelog.pop(key, None)
-        else:
-            self._changelog[key] = (orig, value)
+    def _update_diff(self, key, value):
+        prev = self._data.get(key, MISSING)
+        self._orig.setdefault(key, prev)
+        # If a value changes, dump the cached changelog
+        self._cached_changelog = None
 
     def get_changelog(self):
-        return self._changelog.copy()
+        if self._cached_changelog is None:
+            self._cached_changelog = {
+                key: (old, new)
+                for key, old in six.iteritems(self._orig)
+                for new in [self._data.get(key, MISSING)]
+                if new != old
+            }
+        return self._cached_changelog
 
     def consume_changelog(self):
-        old = self._changelog
-        self._changelog = {}
+        old = self.get_changelog()
+        self._orig = {}
+        self._cached_changelog = {}
         return old
 
     def copy(self):
         new = AssignmentSet()
         new._data = self._data.copy()
-        new._changelog = self._changelog.copy()
+        new._orig = self._orig.copy()
         new._nassigned = self._nassigned
         return new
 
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

**Comment by johntyree** [comment]
TravisCI is expected to fail until #54 is merged.

File: simplesat/dependency_solver.py [hunk 1]
```diff
@@ -52,12 +52,7 @@ def _create_rules(self, request):
             pool.package_id(package)
             for package in pool.what_provides(requirement)
         ]
-        self._policy.add_packages_by_id(requirement_ids)
-
-        # Add installed packages.
```
johntyree: `add_requirements` now implies hard requirements, not just packages we should prefer if possible. The `Policy` gets a reference to `installed_repository` and can figure out on its own which packages are currently installed if it wants to.

simplesat/sat/assignment_set.py
```diff
@@ -0,0 +1,92 @@
...
+    def _update_changelog(self, key, value):
+        if key in self._changelog:
+            orig = self._changelog[key][0]
+        elif key in self._data:
+            orig = self._data[key]
+        else:
+            orig = MISSING
+
+        if orig == value:
+            self._changelog.pop(key, None)
```
johntyree: We should know at this point whether the key is in there or not...

simplesat/sat/policy.py
```diff
@@ -1,39 +1,179 @@
...
+
+    def _log_pretty_pkg_id(self, pkg_id):
+        p = self._log_pool._id_to_package[pkg_id]
+        return '{} {}'.format(p.name, p.version)
+
+
+class _InstalledFirstPolicy(IPolicy):
+
+    CURRENT = count(-2e10)
+    REQUIRED = count(-1e10)
```
johntyree: These should probably be lowered and made `int`.

simplesat/sat/policy.py
```diff
@@ -1,39 +1,179 @@
...
+            self._seen_pkg_ids,
+            key=lambda p: self._pool._id_to_package[p].version,
+            reverse=True,
+        )
+
+        for priority, pkg_id in enumerate(ordered_pkg_ids):
+
+            # Determine the new priority of this pkg
+            if pkg_id in self._required_pkg_ids:
+                new = next(self.REQUIRED)
```
johntyree:
```
self.REQUIRED = int(-1e7)
...
new = priority + self.REQUIRED
```

simplesat/sat/policy.py
```diff
@@ -1,39 +1,179 @@
...
+    def add_requirements(self, package_ids):
+        self._required_pkg_ids.update(package_ids)
+        self._seen_pkg_ids.update(package_ids)
+        self._update_pkg_id_priority()
+
+    def get_next_package_id(self, assignments, clauses):
+        self._update_cache_from_assignments(assignments)
+
+        # Grab the most interesting looking currently unassigned id
+        return self._unassigned_pkg_ids.peek()
```
johntyree: We leave this in queue until we have positive confirmation that it is given an assignment via the changlog.

simplesat/sat/policy.py
```diff
@@ -1,39 +1,179 @@
...
+        self._installed_pkg_ids = set(
+            pool.package_id(package) for package in
+            installed_repository.iter_packages()
+        )
+        self._seen_pkg_ids = set()
+        self._required_pkg_ids = set()
+        self._pkg_id_priority = {}
+        self._unassigned_pkg_ids = PriorityQueue()
+
+    def _update_pkg_id_priority(self, assignments=None):
```
johntyree: It's important that a pkg_id be given the same priority every time it's seen so that it can more or less regain its place in line after being assigne and unassigned repeatedly.

If the population of variables changes though, everyone starts over with a new priority value.

simplesat/sat/policy.py
```diff
@@ -1,39 +1,179 @@
...
+
+    def get_next_package_id(self, assignments, clauses):
+        self._update_cache_from_assignments(assignments)
+
+        # Grab the most interesting looking currently unassigned id
+        return self._unassigned_pkg_ids.peek()
+
+    def _update_cache_from_assignments(self, assignments):
+        for key, (old, new) in six.iteritems(assignments.get_changelog()):
+            if key not in self._seen_pkg_ids:
```
johntyree: This is probably very cheap, but we don't *have* to do it every time. We could track whether we've updated this round or not.

johntyree: `if any(key not in seen for key in changelog): ...`

simplesat/sat/priority_queue.py
```diff
@@ -0,0 +1,99 @@
...
+# -*- coding: utf-8 -*-
+
+from functools import partial
+from heapq import heappush, heappop
+from itertools import count
+
+
+class _REMOVED_TASK(object):
+    pass
+REMOVED_TASK = _REMOVED_TASK()
```
johntyree: `REMOVED_TASK = object()`

johntyree: Or maybe not. I guess it's useful to see something other than `<object>` when debugging.

simplesat/sat/priority_queue.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __contains__(self, key):
+        return key in self._entry_finder
+
+    def push(self, task, priority=0):
+        "Add a new task or update the priority of an existing task"
+        return self._push(priority, self._next_id(), task)
+
+    def peek(self):
+        """ Return the next task, without popping it. """
```
johntyree: Well... we do pop it. But we put it back with the same priority and id so it should go back to the same place. Consider rewording this.

simplesat/sat/priority_queue.py
```diff
@@ -0,0 +1,99 @@
...
+        if n is None:
+            n = len(self)
+        result = []
+        for _ in range(n):
+            result.append(self.pop())
+        return result
+
+    def discard(self, task):
+        "Remove an existing task if present. If not, do nothing."
+        try:
```
johntyree: I think a `__contains__` check is faster here, but I'm not sure we even use this method anyway.

simplesat/sat/tests/test_priority_queue.py
```diff
@@ -0,0 +1,208 @@
...
+
+        pq.discard(1)
+        self.assertTrue(pq)
+
+        pq.remove(0)
+        self.assertFalse(pq)
+
+    def test_tasks_are_unique(self):
+        pq = PriorityQueue()
+        pq.push(0)
```
johntyree: Not sure what happened to the rest of this...

simplesat/sat/assignment_set.py
johntyree: We do but it's ugly and doesn't save us anything to make use of it.

johntyree:
I marked the `iris` test as an expected failure for now.

johntyree:
Failures are due to `AssignmentSet` having an unspecified iteration order. Having looked at all the places where it is used, it's not clear to me why we need that.

Surely the `Policy` should not be relying on whatever order `MiniSatSolver` has decided to put things into the assignment dict (otherwise it's really the `MiniSatSolver` that is doing the deciding) and it appears that everyone else only uses it for lookups.

simplesat/sat/assignment_set.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+
+import six
+
+from collections import OrderedDict
```
cournape: nitpick: let's put stdlib import first

simplesat/sat/assignment_set.py
```diff
@@ -0,0 +1,113 @@
...
+    def __setitem__(self, key, value):
+        assert key > 0
+
+        prev_value = self.get(key)
+
+        if prev_value is not None:
+            self._assigned_literals.difference_update((key, -key))
+
+        if value is not None:
+            self._assigned_literals.add(key if value else -key)
```
cournape: not sure I understand that code: if value is not None, the if inside in the `add` is redundant, right (value cannot be `0` I assume ?)

cournape: Consequently, assuming the above comment is accurate, when does `_assigned_literals` get negative keys ?

simplesat/dependency_solver.py
cournape: from an API POV, I wonder if it would not be simpler to remove `add_packages_by_id` altogether, and make it an implementation detail (called at creation time from a passed `installed_repository`).

simplesat/sat/policy.py
```diff
@@ -1,39 +1,179 @@
...
+        self._required_pkg_ids = set()
+        self._pkg_id_priority = {}
+        self._unassigned_pkg_ids = PriorityQueue()
+
+    def _update_pkg_id_priority(self, assignments=None):
+        pkg_id_priority = {}
+        if assignments is not None:
+            self._seen_pkg_ids.update(assignments.keys())
+
+        ordered_pkg_ids = sorted(
```
cournape: would it be more consistent to call this `ordered_seen_pkg_ids` ?

simplesat/sat/policy.py
```diff
@@ -1,39 +1,179 @@
...
+        pretty = '\n'.join(lines)
+        return c, pretty
+
+    def _log_pretty_pkg_id(self, pkg_id):
+        p = self._log_pool._id_to_package[pkg_id]
+        return '{} {}'.format(p.name, p.version)
+
+
+class _InstalledFirstPolicy(IPolicy):
+
```
cournape: Can we have a small docstring about this class behaviour here ?

simplesat/sat/policy.py
```diff
@@ -1,39 +1,179 @@
...
+        ordered_pkg_ids = sorted(
+            self._seen_pkg_ids,
+            key=lambda p: self._pool._id_to_package[p].version,
+            reverse=True,
+        )
+
+        for priority, pkg_id in enumerate(ordered_pkg_ids):
+
+            # Determine the new priority of this pkg
+            if pkg_id in self._required_pkg_ids:
```
cournape: If I understand correctly the logic, we have one priority queue divided in 3 non-overlapping bands, whose boundaries are defined w/ `REQUIRED` and `CURRENT`. What's the rationale to use this instead of 3 priority queues ?

simplesat/sat/priority_queue.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def pop(self):
+        'Remove and return the lowest priority task. Raise KeyError if empty.'
+        _, _, task = self._pop()
+        return task
+
+    def pop_many(self, n=None):
+        """ Return a list of length n of popped elements. If n is not
+        specified, pop the entire queue. """
+        if n is None:
```
cournape: nitpick: simpler w/ `n = n or len(self)`

cournape:
The PR is a bit big, can we split into at least 2 ?

* the assignment optimisation (that one should be easy to review/accept now)
* the new policy

simplesat/sat/priority_queue.py
mdickinson: Simpler, but less readable, and wrong. :-)  (What if `n = 0`?.)

cournape: you're right, stupid comment :(

mdickinson: Sorry for the drive-by nitpick nitpick.

cournape: It could have been worse, John may have listened to me !

It is nice to know you can count on colleagues to embarrass you :)

PR #56, opened 2015-11-05
Raise error for unsatisfiable requirement

### LINKED ISSUE #47: Fix handling of request with no candidates
Right now, doing a `DependencySolver.solver(Request("something with no candidate))` does not fail as it should.

johntyree:
When the pool cannot provide any packages that satisfy a requirement, we now raise a `egginst.errors.NoPackageFound` exception with a pretty string.

Closes #47

### DIFF
--- a/simplesat/tests/test_scenarios_policy.py
+++ b/simplesat/tests/test_scenarios_policy.py
@@ -64,11 +64,8 @@ class TestNoInstallSet(TestCase, ScenarioTestAssistant):
         self._check_solution("iris.yaml")
 
     def test_no_candidate(self):
-        self.assertRaises(
-            NoPackageFound,
-            self._check_solution,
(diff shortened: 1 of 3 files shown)


simplesat/tests/test_scenarios_policy.py
```diff
@@ -62,6 +63,13 @@ def test_ipython(self):
     def test_iris(self):
         self._check_solution("iris.yaml")
 
+    def test_no_candidate(self):
+        self.assertRaises(
```
cournape: We don't care about pre `2.6` compat, so we should use the context manager form instead:

```
with self.assertRaises(NoPackageFound):
    self._check_solution(...)
```

cournape:
One comment, LGTM otherwise

PR #57, opened 2015-11-05
Use a smart AssignmentSet class to track variable values

johntyree:
This is half of #55 and introduces the `AssignmentSet` along with its tests. The class functions as a drop-in replacement for `OrderedDict`, but is equipped with some extra methods for tracking changes to the values and validating input.

### DIFF
--- a/simplesat/sat/assignment_set.py
+++ b/simplesat/sat/assignment_set.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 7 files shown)


johntyree:
So far this only brings the code that was in #55 over to here. I'm still addressing comments.

johntyree:
I don't think there are any further comments related to the `AssignmentSet`. This is ready for review, @cournape.

cournape:
LGTM

PR #59, opened 2015-11-06
Avoid pruning away packages to be uninstalled.

johntyree:
The way pruning was done previously was incorrectly filtering out all packages marked for uninstallation.

This PR simply marks all currently installed packages as also relevant, so that they won't be pruned.


johntyree:
Somehow this is pointed at the wrong branch.

PR #60, opened 2015-11-06
Don't prune currently installed packages

johntyree:
A remake of #59.

The way pruning was done previously was incorrectly filtering out all packages marked for uninstallation.

This PR simply marks all currently installed packages as also relevant, so that they won't be pruned.

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -32,23 +32,21 @@ class DependencySolver(object):
 
         if solution is False:
             return None
-        else:
-            solution_ids = _solution_to_ids(solution)
 
-            installed_map = set(
-                self._pool.package_id(p)
-                for p in self._installed_repository.iter_packages()
+        solution_ids = _solution_to_ids(solution)
+
+        installed_map = set(
+            self._pool.package_id(p)
+            for p in self._installed_repository.iter_packages()
+        )
+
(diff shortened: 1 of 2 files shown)


johntyree:
This is better, but it still prunes too much. Investigating.

johntyree:
I think this now handles the problems I was running into before.

I consider this a stop gap however, it will likely be best to remove this altogether and rely on the SAT solver itself to actually produce better solutions up front.

simplesat/dependency_solver.py
```diff
@@ -78,40 +79,50 @@ def _create_rules(self, request):
...
-    for pkg_id in pkg_ids:
-        # We pass in `connected` to avoid re-walking a graph we've seen before
-        connected.update(_connected_nodes(pkg_id, neighborfunc, connected))
-
+    should_include = set()
+    for pkg_id in solution_root_ids:
+        # We pass in `should_keep` to avoid re-walking a subgraph
+        nodes = _connected_nodes(pkg_id, neighborfunc, should_include)
+        should_include.update(nodes)
+    assert should_include.issuperset(solution_root_ids)
```
johntyree: This `assert` isn't bringing much value, the solution will always contain the input by construction, however it's very cheap so let's leave it for now.

cournape:
LGTM

PR #61, opened 2015-11-06
First class unsatisfiable scenarios

johntyree:
#### To preemptively handle a merge conflict, this includes #60. Merging that first and then master merging here will clean it up.

This PR introduces the notion of a failed Transaction, which is simply a Transaction containing a single `FailureOperation` that describes why no successful `Transaction` could be constructed.

I'm open to representing this in a different way, but we need *some* way to represent unsolvable problems so that we can pass error messages around and write decent test cases.


johntyree:
This also extends to multiple reasons for failure pretty easily, should that become necessary.

cournape:
I think it would be more natural to represent this as an exception (most of the code would be the same, but the details inside the exception class).

johntyree:
Hmmm... yes that probably makes more sense. I'll take a look.

johntyree:
@cournape Is that more what you have in mind? I haven't push the exception class down into `egginst` yet to avoid complicating things.

simplesat/sat/minisat.py
```diff
@@ -166,7 +173,7 @@ def search(self):
             else:
                 # Conflict!
                 if root_level == self.decision_level:
-                    return False
+                    raise SatisifiabilityError("unknown conflict")
```
johntyree: Eventually, we'll want to attach things here.

cournape:
A few trivial comments, LGTM otherwise

johntyree:
Hmmm yes. I must have typed it incorrectly the first time and then had auto-complete sabotage the rest.

PR #63, opened 2015-11-09
ENH: commandline flag for disabling pruning in scripts/solve.py

johntyree:
It's difficult to validate the behavior of pruning without a way to programmatically disable it. This PR adds a `--no-prune` flag to disable it.


cournape:
LGTM

PR #64, opened 2015-11-09
MAINT: clarify checks for undetermined clauses

johntyree:
This just uses the appropriate method on `set` for what we want to do, which is to see if the clause shares any literals with the set of solution literals. I think it's easier to read.

It should also be slightly faster, assuming these are implemented well.


cournape:
LGTM

PR #65, opened 2015-11-11
A faster InstalledFirst policy (remake of #55)

### LINKED ISSUE #42: Crash with simple `>=` request

johntyree:
This is a remake of the second half of #55.

This PR introduces a Policy that provides "installed first" behavior to the dependency solver, as well as a few helper classes.

Benchmarks are in issue #52.

It is implemented as a giant priority queue where the priorities split into three bands, in order from highest to lowest:

1. Installed packages
  - Packages which are already installed on the system.
2. Required packages
  - Packages which have been explicitly requested.
  - This is lower priority than the installed packages because we don't want to unnecessarily upgrade things.
3. Everything else.

Within each group, packages are ordered descending by version, such that newer packages are tried first.

I'm currently using the notion that version objects are always comparable to define an ordering over *all* packages and assign priorities based on that, rather than grouping packages by name. In addition to being fast, easy to write, and easy to understand, it leads to a few interesting properties:

### Things it does
-  it suggests packages with large version numbers earlier, i.e. `curl-7.3.0` comes well before `Biggus-0.7.0`.
  - This was very easy to implement, but there's no good reason to do it this way. At best it doesn't hurt us. There are plenty of ways to make this smarter.
- it suggests packages interleaved with each other, i.e. `pep8 1.4.6` then `scipy 0.15.0` then `pep8 0.6.1`,
- high priority packages are suggested repeatedly until all options are exhausted,
- not crash on #42,
- it can totally ignore the `clause` literals and still be quite fast.

To see this play out, there's a `PolicyLogger` class that is currently wrapping `InstalledFirstPolicy` and keeping track of things.

### Things it doesn't do
- Anything smart by examining `clause` instances themselves,
- Anything clever with package metadata (i.e. should we suggest packages with few or many dependencies first?)

If this is suitable, I would think that this closes #42 and #52.

### DIFF
--- a/simplesat/sat/policy.py
+++ b/simplesat/sat/policy.py
@@ -87,8 +87,7 @@ class UndeterminedClausePolicy(IPolicy):
     def __init__(self, pool, installed_repository):
         self._pool = pool
         self._installed_map = set(
-            pool.package_id(package) for package in
-            installed_repository.iter_packages()
+            pool.package_id(package) for package in installed_repository
         )
         self._decision_set = self._installed_map.copy()
 
@@ -101,6 +100,12 @@ class UndeterminedClausePolicy(IPolicy):
         """
 
         decision_set = self._decision_set
+        # Remove everything that is currently assigned
+        if len(decision_set) > 0:
+            decision_set.difference_update(
+                a for a, v in six.iteritems(assignments)
+                if v is not None
+            )
         if len(decision_set) == 0:
             decision_set, candidate_id = \
                 self._handle_empty_decision_set(assignments, clauses)
@@ -121,8 +126,6 @@ class UndeterminedClausePolicy(IPolicy):
         assert assignments[candidate_id] is None, \
             "Trying to assign to a variable which is already assigned."
 
-        # Clear out decision set.
-        self._decision_set = set()
         return candidate_id
 
     def _group_packages_by_name(self, decision_set):
--- a/simplesat/sat/policy.py
+++ b/simplesat/sat/policy.py
@@ -1,4 +1,6 @@
 import abc
+from collections import Counter
+from functools import partial
 
 import six
 
@@ -26,6 +28,40 @@ class IPolicy(six.with_metaclass(abc.ABCMeta)):
         clauses : List of Clause
             The collection of Clause objects to satisfy.
         """
+
+
+class PolicyLogger(IPolicy):
+
+    def __init__(self, PolicyFactory, pool, installed_repository):
+        self._policy = PolicyFactory(pool, installed_repository)
+        self._log_installed = list(installed_repository.iter_packages())
+        self._log_suggestions = []
+        self._log_required = []
+        self._log_pool = pool
+        self._log_assignment_changes = []
+
+    def get_next_package_id(self, assignments, clauses):
+        self._log_assignment_changes.append(assignments.get_changelog())
+        pkg_id = self._policy.get_next_package_id(assignments, clauses)
+        self._log_suggestions.append(pkg_id)
+        return pkg_id
+
+    def add_requirements(self, package_ids):
+        self._log_required.extend(package_ids)
+        self._policy.add_requirements(package_ids)
+
+    def _log_histogram(self, pkg_ids):
+        c = Counter(pkg_ids)
+        lines = (
+            "{:>25} {}".format(self._log_pretty_pkg_id(k), v)
+            for k, v in sorted(c.items(), key=lambda p: p[1])
+        )
+        pretty = '\n'.join(lines)
+        return c, pretty
+
+    def _log_pretty_pkg_id(self, pkg_id):
+        p = self._log_pool._id_to_package[pkg_id]
+        return '{} {}'.format(p.name, p.version)
 
 
 class DefaultPolicy(IPolicy):
@@ -238,4 +274,4 @@ class PriorityQueuePolicy(IPolicy):
         assert ours == theirs, "We failed to track variable assignments"
 
 
-InstalledFirstPolicy = UndeterminedClausePolicy
+InstalledFirstPolicy = partial(PolicyLogger, UndeterminedClausePolicy)
--- a/simplesat/sat/policy.py
+++ b/simplesat/sat/policy.py
@@ -144,15 +144,12 @@ class UndeterminedClausePolicy(IPolicy):
     def _handle_empty_decision_set(self, assignments, clauses):
         # TODO inefficient and verbose
 
-        # The assignments with value None
         unassigned_ids = set(
             literal for literal, status in six.iteritems(assignments)
             if status is None
         )
-        # The rest (true and false)
         assigned_ids = set(assignments.keys()) - unassigned_ids
 
-        # Magnitude is literal, sign is Truthiness
         signed_assignments = set()
         for variable in assigned_ids:
             if assignments[variable]:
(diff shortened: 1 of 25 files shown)


johntyree:
I'll address comments from #55 here as well.

johntyree:
From https://github.com/enthought/sat-solver/pull/55#discussion_r43985620
>If I understand correctly the logic, we have one priority queue divided in 3 non-overlapping bands, whose boundaries are defined w/ REQUIRED and CURRENT. What's the rationale to use this instead of 3 priority queues ?

>If it is about getting the priority of an arbitrary pkg id more easily (one instead of 3 function calls), then I think it would be more readable to have a separate class for the low-level priority handling, in which case I don't mind how it is implemented internally (3 bands vs 3 priority queues).

This is indeed to make the code simpler and faster since we are pushing and removing from the priority queue tens of thousands of times for a reasonably sized problem. Given our recent discussion from #62, I will think about how to cheaply split this up to support different behaviors at runtime.

johntyree:
@cournape I found it difficult to split the functionality apart cleanly, but hopefully this is something like what you had in mind :thought_balloon:

johntyree:
So #68 updates the original `InstalledFirstPolicy` in a way that it is competitive with this one and out performs it on certain problems.

Since this PR preserves that policy as `UndeterminedClausePolicy`, I still think this PR is relevant and should be merged. It's also not clear whether *either* policy behaves as outlined in #62 yet, but I believe the priority queue approach is easier to understand in that respect.

simplesat/sat/policy.py
```diff
@@ -82,12 +140,16 @@ def _group_packages_by_name(self, decision_set):
 
     def _handle_empty_decision_set(self, assignments, clauses):
         # TODO inefficient and verbose
+
+        # The assignments with value None
```
cournape: I think this comment is redundant

simplesat/sat/policy.py
```diff
@@ -82,12 +140,16 @@ def _group_packages_by_name(self, decision_set):
...
         unassigned_ids = set(
             literal for literal, status in six.iteritems(assignments)
             if status is None
         )
+        # The rest (true and false)
```
cournape: ditto

simplesat/sat/policy.py
```diff
@@ -82,12 +140,16 @@ def _group_packages_by_name(self, decision_set):
...
         assigned_ids = set(assignments.keys()) - unassigned_ids
 
+        # Magnitude is literal, sign is Truthiness
```
cournape: ditto

simplesat/sat/policy.py
```diff
@@ -104,6 +166,7 @@ def _handle_empty_decision_set(self, assignments, clauses):
             literals = clause.lits
             undecided = unassigned_ids.intersection(literals)
 
+            # The set of relevant literals still undecided
```
cournape: ditto

simplesat/sat/priority_queue.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def __init__(self, order_key_func=lambda x: x):
+        """
+        Parameters
+        ----------
+        `order_key_func` : callable
+            used to sort(reverse=True) items in each group.
+        """
+        self.key_func = order_key_func
+        self._priority_groups = {}
```
cournape: Can't we just set `_priority_groups` as a `defaultdict(set)` ?

simplesat/sat/priority_queue.py
```diff
@@ -0,0 +1,160 @@
...
+    def items(self):
+        "Return an (item, priority) iterator for all items."
+        if self.dirty:
+            self._prioritize()
+        return six.iteritems(self._item_priority)
+
+    def update(self, items, group):
+        "Add `items` to the `group` and update all priority values."
+        self.known = self.known.union(items)
+        for g, pkg_set in self._priority_groups.items():
```
cournape: can we keep the same conventions throughout the class  for `groups` and `item` ? E.g. `for _group, _items in self._priority_groups ...` ?

simplesat/sat/priority_queue.py
```diff
@@ -0,0 +1,160 @@
...
+        # unique id genrator for tie-breaking
+        self._next_id = partial(next, count())
+
+    def __len__(self):
+        return len(self._entry_finder)
+
+    def __bool__(self):
+        return bool(len(self))
+
+    def __contains__(self, key):
```
cournape: can we call key task for consistency ?

simplesat/sat/policy.py
johntyree: These comments are here because it took me quite a while to understand what this policy was doing. For someone who knows the code base, they do not add much value. I'll remove them now.

simplesat/sat/priority_queue.py
johntyree: I wanted to preserve the `KeyError` when an unknown group is accessed. It's perhaps more clear to handle it the other way around.

johntyree:
@cournape ready for re-review.

johntyree:
This PR is getting dismantled and broken into at least two, maybe more.

PR #66, opened 2015-11-12
BUG: UpdateOperation constructor args swapped

johntyree:
The test harness constructs `UpdateOperation` objects backwards.


cournape:
That is is likely on me :(

LGTM

PR #67, opened 2015-11-12
MAINT: clarify _solution_to_ids

johntyree:
Just a style update. I'm not sure if the `+1` and `-1` was intended to make it easier to read, but I've removed it in favor of signed variables.

This should not introduce any functional changes.


cournape:
Not sure what the original rationale was. LGTM

PR #69, opened 2015-11-12
ENH: show dep solver running time in solve script

johntyree:
I found myself reimplementing this constantly. Since the solve script is already dumping everything on stdout, outputting the solving time to stderr seems reasonable to me.


cournape:
LGTM

PR #71, opened 2015-11-13
Support removing packages

### LINKED ISSUE #43: Fix the remove case
We need to be able to remove packages (and the packages that depend on those packages) with confirmation.

Eg. 
1. MKL, numpy and pandas installed
2. Ask to remove MKL
3. Confirm removal of MKL, numpy and pandas.
4. Remove all three.

johntyree:
This PR adds the ability to uninstall packages and their reverse dependencies.

It turns out this was already supported, but for reasons unclear to me extra rules were being generated to force all currently installed packages to remain so. I simply removed the call to create those rules.

Closes #43


cournape:
LGTM

PR #72, opened 2015-11-24
BUG: fix missing import in test case

johntyree:
I'm not sure how this made it through testing, but `expectedFailure` needs to be imported before we use it :)


cournape:
that's indeed weird

cournape:
LGTM

PR #75, opened 2015-12-04
ENH: add support for multiple operations in a request.

cournape:
@johntyree

### DIFF
--- a/simplesat/transaction.py
+++ b/simplesat/transaction.py
@@ -1,23 +1,26 @@
 from collections import OrderedDict
 
+from attr import attr, attributes
+
 from enstaller.new_solver import Requirement
 
 
+@attributes
 class Operation(object):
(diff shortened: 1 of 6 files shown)


simplesat/transaction.py
```diff
@@ -1,23 +1,26 @@
 from collections import OrderedDict
 
+from attr import attr, attributes
```
johntyree: Where does this come from? Is it a package that must be installed?

johntyree:
Code looks reasonable. One question about an import.

simplesat/transaction.py
cournape: Yes, it is attrs: https://pypi.python.org/pypi/attrs

cournape: Let me add it as a dependency, I thought we already had it

PR #77, opened 2015-12-10
Track the search path of the solver

johntyree:
Adds a `PolicyLogger` class which tracks the suggestions of a Policy and the effect that they have on the assignments. This gives us a crude way to mentally retrace the path that the SAT solver took through the problem. It helps with identifying why particular decisions were made along the way, such as which suggested package forced a particular decision and when.

It should have no effect on the results themselves or the outcome of any testing.

### DIFF
--- a/simplesat/sat/policy.py
+++ b/simplesat/sat/policy.py
@@ -34,7 +34,12 @@ class DefaultPolicy(IPolicy):
         return undecided[0]
 
 
-class InstalledFirstPolicy(IPolicy):
+class UndeterminedClausePolicy(IPolicy):
+
+    """ An IPolicy that gathers all undetermined packages from clauses whose
+    truth value is not yet known and suggests them in descending order by
+    package version number. """
+
     def __init__(self, pool, installed_repository):
         self._pool = pool
         self._decision_set = set()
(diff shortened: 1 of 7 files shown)


simplesat/sat/policy.py
```diff
@@ -13,7 +20,99 @@ def get_next_package_id(self, assignments, clauses):
         """
 
 
+class PolicyLogger(IPolicy):
```
cournape: API-wise, I wonder if it would not be simpler to have the logger be a separate class (not implementing `IPolicy`), and create instances within policy classes.

cournape:
Code looks good, but one suggestion regarding API (can be done in separate PR if you prefer).

simplesat/sat/policy.py
johntyree: I'll make a new PR

johntyree: After fiddling around with it, I'm not sure that this is better. By wrapping policies rather than the other way around, we make it easy to catch everything on the way in and out and all of the code related to logging is consolidated into this one class.

cournape: Hm, that's a good point. Can you create a ticket related to this issue, and let's get back to it later while we've worked more on the `*Policy` classes.

PR #78, opened 2015-12-10
Speed up the UndeterminedClausePolicy

johntyree:
This PR was extracted from #65. It tries to improve on the `UndeterminedClausePolicy` by computing as little as possible with each request for a new suggested package and then reusing as much of the result as possible.

It does change the behavior slightly, but since the old behavior was quite wrong anyway, I'm not worried about preserving it.

The goal right now is to get this into a usable state. From there we can worry about nuanced behavior.

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -66,11 +66,6 @@ class DependencySolver(object):
                 raise NoPackageFound(str(requirement), requirement)
             self._policy.add_requirements(requirement_ids)
 
-        # Add installed packages.
-        self._policy.add_requirements(
-            [pool.package_id(package) for package in installed_repository]
-        )
-
         installed_map = collections.OrderedDict()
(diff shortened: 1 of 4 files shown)


cournape:
Can you give some numbers in terms of performance ?

johntyree:
Excellent point. I'll do some tests.

johntyree:
Keep in mind that there's at least 20% random variation here.

### Master
```
ELAPSED : Generate Rules       : 1.102630e+00
ELAPSED : Solver Init          : 2.530599e-01
ELAPSED : SAT Solve            : 4.663004e+00
```

### This branch
```
ELAPSED : Generate Rules       : 1.123396e+00
ELAPSED : Solver Init          : 2.863220e-01
ELAPSED : SAT Solve            : 1.632709e+00
```

cournape:
LGTM

PR #79, opened 2015-12-11
More granular timings

johntyree:
Now we time rule generation, SAT solver initialization, and problem solving separately.


johntyree:
Timings look like https://github.com/enthought/sat-solver/pull/78#issuecomment-163807395 now.

cournape:
LGTM

PR #80, opened 2015-12-11
Show repository_info.name in log

johntyree:


simplesat/sat/policy.py
```diff
@@ -59,7 +59,11 @@ def _log_pretty_pkg_id(self, pkg_id):
         package = self._log_pool._id_to_package[pkg_id]
         name_ver = '{} {}'.format(package.name, package.version)
         fill = '.' if pkg_id % 2 else ''
-        return "{:{fill}<30} {}".format(name_ver, pkg_id, fill=fill)
+        try:
+            repo = package.repository_info.name
```
cournape: In which case did you get an attribute error ?

johntyree: I believe it was possible to have `repository_info` be `None` because of testing. After discussion, I'll go through and fix the tests to generate well-formed packages instead.

johntyree: I can no longer reproduce this at all. I'm not sure why it's there or if it does anything meaningful. I'm going to scrap it.

johntyree:
Removed the `try/except` wrapper. The rest should still be useful, since I want to see the repo in the log.

cournape:
Well, LGTM I guess :)

simplesat/sat/policy.py
johntyree: @cournape, the attribute error is because `InstalledPackageMetadata` has no `repository_info`.
```
  File "/home/john/projects/enthought/jaguar/.devenv-jag34/bin/jaguar", line 9, in <module>
    load_entry_point('jaguar', 'console_scripts', 'jaguar')()
...
...
  File "/home/john/projects/enthought/sat-solver/simplesat/sat/policy.py", line 76, in _log_pretty_pkg_id
    repo = package.repository_info.name
AttributeError: 'InstalledPackageMetadata' object has no attribute 'repository_info'

```

PR #82, opened 2015-12-11
Consolidate scenario

cournape:
@johntyree should remove any reference from `enstaller.new_solver.yaml_utils`. I also backported the tests and one feature from the code in enstaller.

PR #83, opened 2015-12-11
Add marked packages to scenarios

johntyree:
This is based on #82 

This adds support for a `marked` list to the `Scenario` and a few tests for it.

This helps define explicit what behavior should be. If a package is marked then some version of it *must* remain installed. If it is not marked and not involved in a request, then whether or not the package will be removed is unspecified.

### DIFF
--- a/simplesat/test_utils.py
+++ b/simplesat/test_utils.py
@@ -99,16 +99,19 @@ class Scenario(object):
         packages = collections.OrderedDict(
             parse_package_list(data.get("packages", []))
         )
-        operations = data.get("request", [])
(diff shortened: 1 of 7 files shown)


simplesat/test_utils.py
```diff
@@ -100,13 +101,19 @@ def from_yaml(cls, file_or_filename):
         )
         operations = data.get("request", [])
 
+        marked = set(data.get("marked", []))
```
cournape: I would suggest a list for reproducibility

cournape:
LGTM

PR #86, opened 2015-12-16
BUG: fix index_to_yaml script for new enstaller.

cournape:
@johntyree


johntyree:
Where can I find an index on which to test this? There are a lot of metadata-looking files but the relevant ones were sqlite, not json.

cournape:
You can get them as follows for now: `curl -vOL https://api.enthought.com/eggs/rh5-64/index.json`

PR #87, opened 2015-12-16
Add support for package update requests.

johntyree:
This PR adds support for "update" requests for single packages and provides some test scenarios for it.

The behavior is equivalent to explicitly requesting to install the latest version of the package that is being updated.


cournape:
Can we make a new PR rebased on master to make the review easier ?

johntyree:
@cournape see #90

PR #89, opened 2015-12-17
Properly support multiple jobs

johntyree:
A very stupid bug. :sob: 

Now we properly gather up all the requirement ids from *each* job and return those.


cournape:
:blush:

cournape:
LGTM

PR #90, opened 2015-12-17
Add support for package update requests.

johntyree:
Remake of #87.

This PR adds support for "update" requests for single packages and provides some test scenarios for it.

The behavior is equivalent to explicitly requesting to install the latest version of the package that is being updated.

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -1,4 +1,5 @@
 import collections
+from operator import attrgetter
 
 import six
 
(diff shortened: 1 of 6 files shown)


simplesat/dependency_solver.py
```diff
@@ -54,22 +57,28 @@ def solve(self, request):
...
-            if len(requirement_ids) == 0:
+            providers = tuple(pool.what_provides(requirement))
+            if len(providers) == 0:
                 raise NoPackageFound(str(requirement), requirement)
+
+            if job.kind == JobType.update:
+                # An update request *must* install the latest package version
+                providers = [max(providers, key=attrgetter('version'))]
+
+            requirement_ids = list(map(pool.package_id, providers))
```
cournape: can we use list comprehension instead ? It is the recommended python style

simplesat/rules_generator.py
```diff
@@ -288,6 +290,7 @@ def _add_package_rules(self, package):
                         self._add_rule(rule, "package")
 
     def _add_updated_packages_rules(self, package):
+        # FIXME (jtyree 16-12-2015): this looks very much like dead code
```
cournape: indeed, we can safely kill that method here

cournape:
A trivial style comment, some dead code that can be removed, LGTM otherwise

PR #91, opened 2015-12-17
Support preferring new packages over installed packages.

johntyree:
This adds a flag to the policy that allows us to turn off the preference for packages that are already installed on the system. It can be toggled at any time, but the envisioned use case is to select it at initialization and use it throughout.

Additionally a bug is fixed wherein the same version of a package available in the installed repo is chosen from the remote repo. The test scenario provided suffered from this issue prior to this PR.


simplesat/dependency_solver.py
```diff
@@ -16,7 +16,7 @@
 
 class DependencySolver(object):
     def __init__(self, pool, remote_repositories, installed_repository,
-                 policy=None, use_pruning=True):
+                 policy=None, use_pruning=True, prefer_installed=True):
```
cournape: I would prefer putting that choice in the policy class, to avoid polluting the signature of `DependencySolver`.

johntyree: It's just getting passed through here. I'll refactor.

cournape:
Besides the arg comment, I would suggest to add another scenario similar to the `update_reverse_dependencies_no_prefer_installed.yaml` one, but using the prefered installed first policy, to easily compare the solution. Or do we already have such an example ?

johntyree:
`update_reverse_dependencies.yaml` is the same scenario, but with the original solution using the installed first policy.

simplesat/dependency_solver.py
johntyree: :+1:

johntyree:
@cournape I have removed the new kwarg from the dep solver API and added some comments to the two scenario files to indicate how they correspond.

cournape:
LGTM

johntyree:
Somehow this PR is not merged into master.

johntyree:
Scratch that, some of my local refs were stale and reporting no updates somehow. Everything's probably fine here.

PR #92, opened 2015-12-17
Support updating all marked packages to latest version

johntyree:
This adds `update_all` to the Scenario and provides some test cases to show that this works correctly. There isn't really any new logic here. Think of it more as a constructive proof. :bowtie:

TODO:
- [x] add Scenario with conflicting latest-versions of marked packages


cournape:
LGTM

PR #94, opened 2015-12-23
Move pool into sat

cournape:
This migrates the Pool from `enstaller.new_solver` into `simplesat` (should be merged after `maintenance/0.1.x` branch is cut)

### DIFF
--- a/simplesat/test_data/__init__.py
+++ b/simplesat/test_data/__init__.py
@@ -1,0 +1,7 @@
+import os.path
+
+
+HERE = os.path.dirname(__file__)
+
+INDICES_DIR = os.path.join(HERE, "indices")
+NUMPY_INDEX = os.path.join(INDICES_DIR, "numpy_index.json")
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 *.pyc
-*index.json
+./*index.json
 *.egg-info
 .coverage
 .tox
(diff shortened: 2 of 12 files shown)


cournape:
@johntyree first step to remove dependency on enstaller

simplesat/test_data/__init__.py
```diff
@@ -0,0 +1,7 @@
+import os.path
```
johntyree: Why not use `package_data` and `pkgutils` for this?

johntyree:
Should remember to port https://github.com/enthought/enstaller_private/pull/638

simplesat/test_data/__init__.py
cournape: I don't like those packages, they incur too much complexity IMO.

johntyree: You've already added these to the package_data in `setup.py`, though, so it should already "just work."

We're already using that method for the scenario yaml files *and* it claims to be robust against zipped eggs. I also found it pretty easy to use while setting up airspeed-velocity for benchmarks.

Having a second way to compute (sometimes correct) paths won't interfere with that, but using this requires that eggs be unzipped, doesn't it?

johntyree:
I think this is alright as-is. I'll just complain about not using `pkgutil.get_data` in future PRs. :wink:

johntyree:
Oh hmm. Meant to do this after making the maintenance branch...

I suppose I'll just have to make that branch from the parent of this.

PR #95, opened 2015-12-23
DOC: create initial changelog. Changes for 0.1.0

johntyree:
@cournape

More detail? Less detail?


cournape:
Before the feature section, I would just mention that it supports only runtime dependencies, w/ enpkg-like constraints, but otherwise LGTM

johntyree:
Updates updated.

cournape:
:+1:

PR #97, opened 2015-12-23
BUG: deterministically iterate over package dependencies

### LINKED ISSUE #96: Solving is non-deterministic.
Changing the `PYTHONHASHSEED` can cause the solver to produce different solutions (or failures). This is bad.

johntyree:
I believe this to be the root of the problem.

Closes #96.


rkern:
LGTM.

PR #98, opened 2015-12-25
BUG: prefer installed packages over remote for same version

johntyree:
Without this check we get stuck in infinite `update-all` loops.

It would probably be nice to refactor this somehow and remove the redundancy, but since the rule generation is currently independent from policy initialization, it won't be a superficial change.

### DIFF
--- a/simplesat/tests/test_rules_generator.py
+++ b/simplesat/tests/test_rules_generator.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import io
+import unittest
+
+from enstaller.new_solver import Pool
+
+from ..rules_generator import RuleType, RulesGenerator
+from ..test_utils import Scenario
+
+
+class TestRulesGenerator(unittest.TestCase):
+
+    def test_prefer_installed(self):
+
+        self.yaml = """
+            packages:
+                - A 1.0.0-1
+            installed:
+                - A 1.0.0-1
+            marked:
+                - A
+            request:
+                - operation: "update_all"
+        """
+
+        scenario = Scenario.from_yaml(io.StringIO(self.yaml))
+        repos = list(scenario.remote_repositories)
+        repos.append(scenario.installed_repository)
+        pool = Pool(repos)
+        installed_map = {
+            pool.package_id(p): p for p in scenario.installed_repository}
+        rules_generator = RulesGenerator(pool, scenario.request, installed_map)
+        rules = list(rules_generator.iter_rules())
+
+        self.assertEqual(len(rules), 2)
+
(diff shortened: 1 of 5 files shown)


cournape:
Can we have a test for this ?

johntyree:
Test added.

simplesat/tests/test_rules_generator.py
```diff
@@ -0,0 +1,48 @@
...
+
+from ..rules_generator import RuleType, RulesGenerator
+from ..test_utils import Scenario
+
+
+class TestRulesGenerator(unittest.TestCase):
+
+    def test_prefer_installed(self):
+
+        self.yaml = u"""
```
cournape: this does not need to be an attribute, right ?

simplesat/tests/test_rules_generator.py
```diff
@@ -0,0 +1,48 @@
...
+
+        update = updates[0]
+
+        self.assertEqual(update.reason, RuleType.job_update)
+        self.assertEqual(len(update.literals), 1)
+
+        pkg_id = update.literals[0]
+        package = pool._id_to_package[pkg_id]
+
+        self.assertEqual(package.repository_info.name, "installed")
```
cournape: We should avoid depending on the repository name. It would be better to check the package is the same as the one you get from the installed repository

simplesat/tests/test_rules_generator.py
```diff
@@ -0,0 +1,48 @@
...
+from enstaller.new_solver import Pool
+
+from ..rules_generator import RuleType, RulesGenerator
+from ..test_utils import Scenario
+
+
+class TestRulesGenerator(unittest.TestCase):
+
+    def test_prefer_installed(self):
+
```
cournape: We generally follow the convention of separating a test into `Given`, `When` and `Then` sections (delimited by comments, see other tests).

simplesat/tests/test_rules_generator.py
johntyree: Nope!

simplesat/tests/test_rules_generator.py
johntyree: I was trying to think of a good way to do this. Of course comparing the packages makes much more sense. Is there a reason not to `assertIs` for this?

simplesat/tests/test_rules_generator.py
johntyree: At first I was going to whine, but when I rewrote it with sections it became much nicer. :grinning:

cournape: It is one of the rare use of comments I find useful :)

cournape:
LGTM

PR #99, opened 2015-12-27
BUG: AssignmentSet may not have seen a package_id

johntyree:
The package_ids we consider come from the Clause literals,
but we look them up in the AssignmentSet. If the
AssignmentSet has never seen that particular id before, it
crashes. This fixes that.

### DIFF
--- a/simplesat/sat/policy.py
+++ b/simplesat/sat/policy.py
@@ -190,7 +190,7 @@ class UndeterminedClausePolicy(IPolicy):
                     assignments
                 )
 
-        assert assignments.get(candidate_id) is None, \
+        assert assignments.value(candidate_id) is None, \
             "Trying to assign to a variable which is already assigned."
 
         if not self.prefer_installed:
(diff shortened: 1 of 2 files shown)


cournape:
I wonder if it makes sense to handle this at the `AssignementSet` level ?

johntyree:
Hmm...

Yes it probably does, given that `AssignmentSet.value` returns `None` for unseen ids.

johntyree:
Turns out this was a bug anyway. Bug is fixed and test added to make sure `get()` and `__getitem__` do what is expected of them.

simplesat/sat/policy.py
```diff
@@ -190,7 +190,7 @@ def get_next_package_id(self, assignments, clauses):
                     assignments
                 )
 
-        assert assignments[candidate_id] is None, \
+        assert assignments.value(candidate_id) is None, \
```
cournape: Why don't we simply use the `get` method as in the tests ?

johntyree: Hmmm. Yes you're right. I had the relationship between ids and literals backwards. The keys in `AssignmentSet` are ids and thus always positive. `value` is used for looking up *literals*. Here we're looking up an id and should use `get`.

PR #100, opened 2015-12-29
ENH: don't wrap long lines in yaml output

johntyree:
Have hard-wrapped lines makes it annoying to manipulate the output in an editor. This prevents the wrapping of the lines.

I wasn't sure how to achieve this in a nice way so just used a very large value. If you know how, do tell.


cournape:
I don't know a better way. :+1:

PR #101, opened 2015-12-29
Show an error message on failure.

johntyree:
This is the first iteration of error reporting for failed scenarios. Most of the effort has been spent trying to make sure we can quickly grab references to the important conflicting clauses as well as the rules and requirements associated with them.

The error is represented as a new class `UNSAT`, which carries along with it the conflicting clause, the set of most recent literal assignments and the clauses that forced them, some helper functions to try to extract information from those clauses and a "default" informative string building function.

No work has been done to make any of it read nicely.


johntyree:
This merges cleanly with master, but will need to be updated after #113 is in.

johntyree:
I don't think there's anything else that needs to be done here. Ready for review. @cournape

johntyree:
@cournape can we make a call on this before it causes more conflicts?

It's also annoying to test things without it.

cournape:
Agreed, I will take a look at it later tonight.

simplesat/sat/minisat.py
```diff
@@ -60,21 +167,22 @@ def __init__(self, policy=None):
...
+        self.assigning_clause = {}
+        self._assignment_seq = count()
 
         # Whether the system is satisfiable.
         self.status = None
 
         self._policy = policy or DefaultPolicy()
 
-    def add_clause(self, clause):
+    def add_clause(self, clause, rule=None):
```
cournape: Can we add a docstring to document `rule` (and `cause` in the `enqueue` method)

cournape:
A few trivial comments, but I am not sure I have a good picture of how this works. We can talk about it during the weekly meeting.

simplesat/sat/minisat.py
johntyree: :+1:

johntyree:
Unless the PR lines are each modified by a single commit, reviewing commit-by-commit is not very useful.

johntyree:
Somehow we managed to forget to talk about this at the meeting. Do you still have questions about how it works?

simplesat/sat/minisat.py
```diff
@@ -12,6 +13,155 @@
...
+        Create a new UNSAT object.
+
+        Parameters
+        ----------
+        conflict : Clause
+            The clause which has been found to be unsatisfiable.
+        learned_clause : Clause
+            The implied change needed to satisfiy the conflict.
+
+            This clause is always a learned clause and should always have
```
cournape: why don't we store the conflicting literal then ?

cournape:
@johntyree the code LGTM, but it needs a few tests

simplesat/sat/minisat.py
johntyree: We store the entire clause because it includes history information that we use to search for the requirement that caused it.

johntyree:
@cournape I wasn't sure about the best way to test the UNSAT object without recreating an entire scenario, so what I've done instead is made the scenario files specify exactly how they fail. Now for an unsatisfiable scenario to pass, the UNSAT object must have the correct requirements in the correct order and must produce an exactly matched string.

The exact string match might not be useful while we're still hammering out how error messages will be dealt with, particularly since it's based on the default `__str__` of the conflicting rules and requirements, but this will at least raise a red flag every time something happens that changes it.

cournape:
I think that method is fine for now I think.

cournape:
LGTM

PR #104, opened 2015-12-31
REF: move Requirement from enstaller into simplesat.

cournape:
Built on top of #94 , so should be merged after

PR #105, opened 2015-12-31
Move requirement into sat 2

johntyree:
This is just fixing the dangling commits from https://github.com/enthought/sat-solver/pull/104 and updating a few more imports.


johntyree:
Merging this, as it was only omitted from #94 due to error.

PR #106, opened 2016-01-01
Add benchmarks using airspeed-velocity

johntyree:
This PR introduces a dev dependency on the current master over at https://github.com/spacetelescope/asv and adds an appropriate config file for running benchmarks with the `asv` command.

Some points to note:
- `asv` doesn't seem to support requirements files, so I wrote a small plugin to do so.
- The last actual release was 0.1.1, but the API has changed quite a lot since then so I'm using master as of today instead.
- There was some refactoring of `DependencySolver` done to make it easier to benchmark little pieces of it without duplicating code in the benchmarks.
- The only benchmarks currently just walk through the process from yaml to transaction. There apparently is no way to report intermediate times, so the entire process becomes O(n²). Setup for stage 2 is to run stage 1 (again). Setup for stage 3 is to run 1 and 2, etc.


johntyree:
Good news and bad news.

The good news is that asv does a pretty good job when it works.

The bad news is that it has some really silly design choices that make it difficult to develop with. For example, it always discovers benchmarks to run based on the master branch, even if you aren't using the master branch for anything and explicitly specify a different branch in the config file.

The benchmarks in this branch depend on a slight refactoring of `dependency_solver.py`, which means asv can't run them any more because it uses `master` when discovering, causing import errors. I don't have a good answer for this and don't really want to spend time rewriting asv to behave better.

johntyree:
I *suspect* that merging this will cause the benchmarks to suddenly work, but it's pretty annoying.

johntyree:
It looks like the standard way to use asv is to keep a separate repo for benchmarks and results. I'll move this out.

PR #109, opened 2016-01-02
Move constraints and request into sat

cournape:
Move the remaining of `enstaller.new_solver` into `simplesat`.

@johntyree


johntyree:
`_RawConstraintsParser.parse` runs the constraints through a `set` object.

johntyree:
Looks good!

PR #110, opened 2016-01-03
Move repository into sat

cournape:
This PR implements a simplified `Repository` class to replace the enstaller one in simplesat:

1. It works w/ any enstaller `PackageMetadata` subclass
2. It only support the `find_package*` and `update` methods. 
3. Internally, the instance keeps a sorted list of package names to allow reproducible iteration over a repository, independently of the order in which packages have been added.

This is the penultimate PR to remove the dependency on enstaller. The last one will implement a leaner `PackageMetadata` class.

@johntyree


simplesat/repository.py
```diff
@@ -0,0 +1,110 @@
...
+        If the same package is added multiple times to a repository, every copy
+        will be available when calling find_package or when iterating.
+        """
+        if package_metadata.name not in self._name_to_packages:
+            bisect.insort(self._names, package_metadata.name)
+
+        self._name_to_packages[package_metadata.name].append(package_metadata)
+        # Fixme: this should not be that costly as long as we don't have
+        # many versions for a given package.
+        self._name_to_packages[package_metadata.name].sort(
```
sjagoe: What is the argument against making PackageMetadata orderable and using `insort` here too?

sjagoe: Looking at the definition of PackageMetadata, the sorting would occur on `(name, version, dependencies, python_tag)`, which I think would be acceptable here.

sjagoe:
One question, otherwise LGTM

simplesat/repository.py
```diff
@@ -0,0 +1,110 @@
...
+import bisect
+import collections
+import operator
+
+from .errors import NoPackageFound
+
+
+class Repository(object):
+    """
+    A Repository is a set of package, and knows about which package it
```
johntyree: "packages"

simplesat/repository.py
```diff
@@ -0,0 +1,110 @@
...
+        # over a repository reproducible
+        self._names = []
+
+        packages = packages or []
+        for package in packages:
+            self.add_package(package)
+
+    def __len__(self):
+        return sum(
+            len(self._name_to_packages[p])
```
johntyree: What about `.values()`?

simplesat/repository.py
```diff
@@ -0,0 +1,110 @@
...
+        sorted from lowest to highest version.
+
+        Parameters
+        ----------
+        name : str
+            The package's name
+
+        Returns
+        -------
+        packages : iterable
```
johntyree: We already specifically call the return value a list, above. I would change either here or there to make it clear whether or not the type is guaranteed.

simplesat/repository.py
```diff
@@ -0,0 +1,110 @@
...
+            Iterable of RemotePackageMetadata-like (order is from lower to
+            higher version)
+        """
+        if name in self._name_to_packages:
+            return list(self._name_to_packages[name])
+        else:
+            return []
+
+    def update(self, iterable):
```
johntyree: We call it `packages` in the constructor.

simplesat/rules_generator.py
```diff
@@ -1,9 +1,8 @@
 import collections
 import enum
 
-from enstaller.errors import EnstallerException
-
 from .constraints import Requirement
+from .errors import NoPackageFound, SolverException
```
johntyree: Does this trip up higher level tools that would be using this in place of enstaller? They will be expecting the other class object with this name.

Not sat solver's fault, but now we'll have to import and catch both sets of exception classes, right?

simplesat/tests/test_repository.py
```diff
@@ -0,0 +1,170 @@
...
+        """)
+        packages = self.packages_from_definition(packages_definition)
+        repository = Repository(packages)
+
+        # When/Then
+        with self.assertRaises(NoPackageFound):
+            repository.find_package("nose", V("1.4.0-1"))
+
+        with self.assertRaises(NoPackageFound):
+            repository.find_package("nono", V("1.4.0-1"))
```
johntyree: I think it would be better to choose a version number that exists so that it's clearly the name causing failure.

simplesat/repository.py
cournape: I think I benchmarked that for small sequences, sorting every time is faster than calling insort, but let me check again since a lot of the code changed since.

cournape: Ok, so there is no measurable speed change in my limited test, but to work properly, it requires a change in `PackageMetadata` (as they are not orderable ATM). Since I will rewrite `PackageMetadata` to be in SAT, I would suggest to make that change then

simplesat/rules_generator.py
cournape: Once `sat-salver` does not depend on `enstaller` anymore, I will move the legacy solver in `sat-solver` as well, so there will be no duplicate. Until then, there is a bit of a chicken and egg situation :)

sjagoe:
LGTM

PR #111, opened 2016-01-07
Dont depend index

cournape:
On top of #110

This removes another API use from enstaller.

@johntyree @sjagoe


sjagoe:
LGTM, but needs to wait for #110 to be reviewed and merged.

PR #114, opened 2016-01-13
Add pretty string for new PackageMetadata

johntyree:
The new classes don't have a nice representation, which makes the output of `scripts/solve.py` pretty unreadable. This just implements `__repr__` on those classes in a way similar to the original.

### DIFF
--- a/simplesat/package.py
+++ b/simplesat/package.py
@@ -88,7 +88,7 @@ class PackageMetadata(object):
 class RepositoryPackageMetadata(object):
     @classmethod
     def _from_pretty_string(cls, s, repository_info):
-        package = PackageMetadata._from_pretty_string(s) 
+        package = PackageMetadata._from_pretty_string(s)
         return cls(package, repository_info)
 
     def __init__(self, package, repository_info):
--- a/simplesat/package.py
+++ b/simplesat/package.py
@@ -29,6 +29,9 @@ class RepositoryInfo(IRepositoryInfo):
     def __ne__(self, other):
         return self._key != other._key
 
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


simplesat/package.py
```diff
@@ -68,6 +71,10 @@ def version(self):
     def dependencies(self):
         return self._dependencies
 
+    def __repr__(self):
+        return "PackageMetadata('{0}-{1}', key={2!r})".format(
+            self._name, self._version, self._key)
```
cournape: I don't think we should print the key in the repr, as it really just an implementation detail (what defines identity, essentially)

cournape:
One comment, LGTM otherwise

simplesat/package.py
```diff
@@ -107,6 +114,12 @@ def dependencies(self):
     def repository_info(self):
         return self._repository_info
 
+    def __repr__(self):
+        return (
+            "RepositoryPackageMetadata('{pkg._name}-{pkg._version}'"
+            ", repo={repository_info!r}".format(
```
cournape: a closing ')' is missing

simplesat/package.py
johntyree: Fair enough. I was just bringing over the code that was already in enstaller.

simplesat/package.py
johntyree: :+1:

simplesat/package.py
cournape: Key in the original code had a completely different meaning (was the egg filename, keeping the casing information lost in the name)

johntyree: Aha. That makes sense.

Issue #108, opened 2016-01-01
Fix the behavior of `==` for upstream version match


cournape:
`~=` has a different meaning in [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-specifiers), so we should use a different syntax to avoid clashes. I would suggest something like `^=` or `#=`.

We should also change `==` to be consistent w/ PEP 440.
See alo [PEP 508](https://www.python.org/dev/peps/pep-0508/)

cournape:
Fixed in #115

johntyree:
We haven't fixed the behaviour of `==` yet, if that's necessary.

cournape:
Indeed, we should fix that as well

cournape:
Tracked more generally in #174

PR #115, opened 2016-01-13
Change from ~= to ^=

johntyree:
This is a pretty brain-dead substitution of the thousands of occurrences of `~=` that appear.

It doesn't change the behavior of `==` according to PEP440. That will be another PR.

xref: https://github.com/enthought/sat-solver/issues/108


cournape:
I did not mention what to replace it with to avoid bike shedding, but I was thinking about using exactly the same symbol :)

LGTM

johntyree:
You did, in fact, mention it. :)

I would probably have gone with `.=`, but I was worried about the dot
causing problems. Anyway, I think this one is good too. No shedding needed.

PR #116, opened 2016-01-14
ENH: 'export' public API to be used by jaguar/enstaller.

cournape:
@johntyree a first step toward fixing jaguar/enstaller/sat-solver integration


johntyree:
LGTM

PR #120, opened 2016-01-19
Change PackageMetdata.dependencies to .install_requires

johntyree:
The word "dependency" still appears throughout the code, but only in local contexts where I think it's clear what it means.


cournape:
LGTM

Issue #118, opened 2016-01-16
Refactor package metadata handling


johntyree:
Before #93 can be put together, we should deal with the following. This roughly describes my plan for documenting and cleaning up this part of the code.

- [x] Remove usages of old-style "legacy" (`numpy 1.8.0`) constraint strings
- [x] Use `install_requires` rather than `dependencies` (c.f. #117)
-  ~~Use `distribution` rather than `name`.~~
- [x] Extend package string parsing to support handling other kinds of constraint blocks
- [x] Discuss and decide on whether a better (non-string) representation of package metadata is needed.

The flow will be as follows.

1. A "pretty" string such as `numpy 1.8.0; install_requires (MKL ^= 10.3)` will be converted to `PackageMetadata(distribution='numpy', version=EnpkgVersion.from_string('1.8.0'), install_requires=("MKL ^= 10.3",))`. Note that everything other than `version` is a string here.
2. `Requirement` objects will be built by new-style parsing the `install_requires` above.
3. `Rule` objects will be created from the `Requirement` objects in the same way they currently are.

johntyree:
Closed by #120 and #122.

PR #122, opened 2016-01-19
Introduce new PackageMetadata format.

johntyree:
This introduces a new format for package metadata, moving us away from using "pretty strings" as our canonical representation.

An excerpt from the PackageMetadata docstring is helpful.
```
install_requires : tuple(tuple(str, tuple(tuple(str))))
    A tuple of tuples mapping distribution names to disjunctions of
    conjunctions of version constraints.

    For example, a consider a package that depends on the following:
        - nose
        - six (> 1.2, <= 1.2.3), or >= 1.2.5-2
            Written as intervals, (1.2, 1.2.3] or [1.2.5-2, \infty)
        - MKL >= 10.1, < 11

    The constraint tuple representing this would be:
        (("MKL", ((">= 10.1", "< 11"),)),
         ("nose", (("",),)),
         ("six", (("> 1.2", "<= 1.2.3"), (">= 1.2.5-2",)))
```

It also includes an example file, which shows the format and will be used in tests soon.
```
{
    "name": "pandas",
    "version": "1.17.0-2",
    "install_requires": {
        "numpy": [["^= 1.8.1"]],
        "scipy": [["> 2"]],
        "click": [["< 6"], [">= 7"]]
    },
    "conflicts": {
        "paneldata": []
    }
}
```

### TODO:
- [ ] Significantly more testing around parsing pretty strings into the new format.
- [ ] Test creating JSON from a `PackageMetadata` object.
- [ ] Implement a new constructor for `Requirement` that can handle a constraint tuple.


cournape:
Is this WIP ?

johntyree:
Sorry, yes mostly. Forgot to edit the title.

johntyree:
This probably has some oddities in it, but I think it's ready for some scrutiny now.

@cournape :bell:

cournape:
Can we create a new PR by rebase after #120 is merged ?

johntyree:
That OrderedDict isn't necessary anymore... I'll revert.

simplesat/constraints/package_parser.py
```diff
@@ -1,69 +1,81 @@
...
-from .parser import _RawRequirementParser
+from .parser import _DISTRIBUTION_R, _VERSION_R, _WS_R
 
 
 DEPENDS_RE = re.compile("depends\s*\((.*)\)")
+CONFLICTS_RE = re.compile("conflicts\s*\((.*)\)")
 
+_MAYBE_WS_R = " *"
+_DISTRIBUTION_R = "(?P<distribution>{})".format(_DISTRIBUTION_R)
+_VERSION_R = "(?P<version>{})".format(_VERSION_R)
```
cournape: Can we have a new name for the new regex ?

simplesat/constraints/package_parser.py
```diff
@@ -1,69 +1,81 @@
...
+            kind = CONSTRAINT_SYNONYMS.get(kind, kind)
+            constraints_str = match.group('constraints')
+            if kind not in VALID_CONSTRAINT_KINDS:
+                msg = "Invalid package string. Unknown constraint kind: {!r}"
+                raise ValueError(msg.format(kind))
+            constraints = {}
+            for match in CONSTRAINT_RC.finditer(constraints_str):
+                dist = match.group('distribution')
+                constraint_str = match.group('constraint')
+                constraints.setdefault(dist, [[]])[0].append(constraint_str)
```
cournape: can we use a `defaultdict` instead ?

simplesat/constraints/requirement.py
```diff
@@ -47,6 +48,41 @@ class Requirement(object):
...
+    @classmethod
+    def from_constraints(cls, constraint_tuple):
+        """ Return a Requirement object from a PackageMetadata constraint
+        tuple.
+
+        Parameters
+        ----------
+        constraints : constraints tuple
+            A tuple of constraints like (
+                'nose', ( # disjuncitons
```
cournape: typo in disjunctions

cournape: What's the rationale to use `("nose", ((">= 1.3", "< 1.4"),))` instead of `{"nose": ((">= 1.3", "< 1.4"),)} ?

simplesat/constraints/requirement.py
```diff
@@ -47,6 +48,41 @@ class Requirement(object):
...
+                    ('=> 1.3', '< 1.4'),  # conjunction
```
cournape: typo: should be `>=`, no `=>`

simplesat/constraints/requirement.py
cournape: Answering to myself: immutability (I guess)

simplesat/constraints/package_parser.py
johntyree: Oof. Yes definitely.

simplesat/constraints/package_parser.py
johntyree: Done

simplesat/constraints/requirement.py
johntyree: Determinism, not immutability.

It has to be done anyway for hashing to work and for things like pretty string printing to be deterministic. In JSON it can be an object instead of array, that's fine. In Python there's no particular reason to use a dict. We don't do arbitrary look-ups and we don't display it to anyone.

simplesat/package.py
```diff
@@ -52,9 +56,33 @@ def _from_pretty_string(cls, s):
...
+            The constraint tuple representing this would be:
+                (("MKL", ((">= 10.1", "< 11"),)),
+                 ("nose", (("",),)),
+                 ("six", (("> 1.2", "<= 1.2.3"), (">= 1.2.5-2",)))
+
+        """
         self._name = name
         self._version = version
-        self._install_requires = install_requires or tuple()
+        self._install_requires = tuple(sorted(install_requires or ()))
```
cournape: I am not sure we want to sort here. Are there API usecases where it is annoying to not sort ?

simplesat/constraints/requirement.py
cournape: You need immutability for hashing :)

johntyree: What's written here is backwards actually. They get sorted so it should be `('< 1.4', '>= 1.3')`.

simplesat/constraints/requirement.py
johntyree: Fixed.

cournape:
Regarding the example, I think this would be simpler no ?

```
install_requires : tuple(tuple(str, tuple(tuple(str))))
    A tuple of tuples mapping distribution names to disjunctions of
    conjunctions of version constraints.

    For example, a consider a package that depends on the following:
        - nose
        - six (> 1.2, <= 1.2.3), or >= 1.2.5-2
            Written as intervals, (1.2, 1.2.3] or [1.2.5-2, \infty)
        - MKL >= 10.1, < 11

    The constraint tuple representing this would be:
        (("MKL", ((">= 10.1", "< 11"),)),
         ("nose", ((,),)),
         ("six", (("> 1.2", "<= 1.2.3"), (">= 1.2.5-2",)))
```

(i.e allowing each level of constraint to be empty)

simplesat/package.py
johntyree: Well considering that the dependencies should definitely not be ordered, this keeps it deterministic without the caller having to know that we rely on that. Is it overkill?

cournape: It is still deterministic as long as don't change the order ;)

I think it is fine to keep things in the same order as in the original string, it is less surprising (at least for me)

johntyree:
Your code isn't valid Python, so I assume you meant `("nose", ((),))`. I had that first actually. Even going so far as `("nose", ())`. What I don't like about it is that it's very easy to swallow everything entirely if you iterate over the group in the most obvious and idiomatic way.

For example:
```
>>> deps
(('MKL', (('>= 10.1', '< 11'),)), ('nose', ((),)), ('six', (('> 1.2', '<= 1.2.3'), ('>= 1.2.5-2',))))
>>> for dist, disjuncts in deps:                                                                                                                                   
...   for conjuncts in disjuncts:
...     for constraint in conjuncts:
...        print(dist + ' ' + constraint)
... 
MKL >= 10.1
MKL < 11
six > 1.2
six <= 1.2.3
six >= 1.2.5-2
```

If we had a proper Option/Maybe type it wouldn't matter so much, but as-is it's just too easy to get wrong for my taste.

cournape:
Ah, I always thought that `()` could not be used to define empty tuples :(

As for the looping argument, I am not terribly convinced because I would consider it more natural to write to print the dist in the outer loop. If you absolutely want to have a value in there, then it should be `"*"`, not `""`.

johntyree:
For printing it makes sense. For things like flattening in to a list to generate requirements, less so.

I'll change it to `"*"` though. I like that more.

johntyree:
I believe the parser does accept both `""` and `"*"`. Do we want to change that?

cournape:
I think it is ok to allow both `""` and `"*"` in pretty strings (e.g. `numpy 1.9.2-1; depends (nose, MKL ^= 10.3)` and `numpy 1.9.2-1; depends (nose *, MKL ^= 10.3)`), but we don't want to allow empty string in the json format.

simplesat/package.py
johntyree: OK. I'm still sorting by distribution name, but not by the `<operator><version>` strings. How about that?

johntyree:
Closing this will finish up https://github.com/enthought/sat-solver/issues/118.

simplesat/package.py
cournape: I was about to say I would rather keep it as, but we actually have to sort for determinism since we keep things in json (and json dict like python ones are not ordered), so I am fine with that

cournape:
LGTM

simplesat/package.py
johntyree: I'm not sure I follow. Can you give an example that demonstrates which part you're thinking needs to be sorted and which part doesn't? Right now, the code sorts based on the names of the reverse-dependencies, but not the constraints associated with them.

For example, the way it currently is we get the following behavior.

```
>>> sorted("><")
['<', '>']

>>> pkg = ppsp.parse("numpy 1.8.0-2; depends (zope > 2.0, bokeh, zope < 1.0)")
>>> pkg['install_requires']
(('bokeh', (('',),)), ('zope', (('> 2.0', '< 1.0'),)))
```

So the order of the package names gets sorted, but the order of the constraint
strings does not. This sounds like what you were suggesting, but can you
confirm?

I just cleaned this up a little bit as well.

johntyree:
Sorry this needs a tiny re-review. Just that last MAINT commit and the comment on the diff above.

cournape:
LGTM

PR #123, opened 2016-01-21
Add support for explicit package conflicts.

johntyree:
This adds explicit package conflicts to scenario files and rule generation. :dancer:

The format is just as with dependencies and it fully supports inequalities, e.g.

```
browsey 1.2.0-3;
    depends (click >= 1.1);
    conflicts (click <= 2.0, click > 1.3, zope ^= 0.19.3)
```

It includes unit tests and a test scenario file. If there's another conflicting scenario that should be included in here, please share it.

@cournape


johntyree:
Note to self: Make sure that for conflicting packages, the removal occurs before the installation.

johntyree:
Thinking about this, how consistent are we trying to be? Do we care about the intermediate states other than for conflicting packages?

Is it enough to always do all removals before all installations?

johntyree:
We could be smarter about it by, say, finding all groups of conflicting packages and only enforcing the "remove before install" rule within the groups, but why bother?

cournape:
@johntyree if possible, it is better to install dependencies first.

Right now, it is not a big deal, but later we may have some packages whose pre/post install scripts may depend on their dependencies. I would like to keep that option available, unless it is too complicated.

johntyree:
Sure, we can do that. I just mean splitting things like package updates into a removal and later installation.

I already have the toposort code from the priority queue branch. I'll just try it out and see.

johntyree:
I think if you want to maintain the invariant that we never try to install or remove a package without its dependencies being installed then we have to do away with the notion of "update" internally.

Currently we have a lot of things that differentiate somehow between "Remove a-0, install a-1" vs "update a from 0 to 1," but if we need to have b-0 installed while a-0 is being uninstalled, and b-1 installed while a-1 is being installed, then we have to split that operation up. From a UI perspective, I understand wanting to know which packages are being updated, but internally I think it should be simplified such that there are only `install` and `remove` operations.

The code for this is pretty straight forward. I have a working implementation already, that is contained entirely with the `Transaction` object. We should probably discuss a bit about how to go forward.

johntyree:
There will be a lot of failing tests because of the removal of "update".

simplesat/transaction.py
```diff
@@ -29,16 +30,20 @@ class RemoveOperation(Operation):
 class Transaction(object):
 
     def __init__(self, pool, decisions, installed_map):
-        self.operations = []
-
+        self.operations = self._safe_operations(pool, decisions, installed_map)
+        self.pretty_operations = []
```
johntyree: The other code is still here for easy comparison. If we decide to do things the new way it should be removed.

johntyree:
@cournape We can talk about this at our Monday meeting. Short of updating the tests, this is good to go.

We should decide if we want to keep the pretty representation with explicit `update` operations to show users. I'm leaning towards "no," so that we only show the user what we actually plan to do, but it depends on whether or not you think we're showing things "in order" or not in the first place.

If we say "update numpy, remove scipy, install seaborn" does that imply that we're going to do those things in that order? If so, then I propose we dump updates entirely. If not, then maybe we can keep them.

simplesat/tests/pil_pillow_upgrade_conflict.yaml
```diff
@@ -0,0 +1,39 @@
+packages:
```
cournape: Are the PIL/Pillow-related scenario used anywhere ?

simplesat/transaction.py
```diff
@@ -73,6 +78,44 @@ def to_simple_string(self):
...
+        for group in toposort(graph):
+            # Sort the set of independent packages for determinism
+            for package_id in sorted(group):
+                assert package_id in decisions
+                if package_id < 0 and -package_id in installed_map:
+                    removals.append(-package_id)
+                elif package_id > 0 and package_id not in installed_map:
+                    installs.append(package_id)
+
+        # Removals should happen top down
```
cournape: Can we have some explicit unit test for this feature ?

simplesat/transaction.py
```diff
@@ -73,6 +78,44 @@ def to_simple_string(self):
 
         return "\n".join(lines)
 
+    def install(self, package):
```
cournape: This was there before this PR, but I don't think install/remove/update should be public APIs. Can we rename them to `_*` ?

simplesat/utils/graph.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
```
cournape: same here, those needs tests :)

simplesat/transaction.py
cournape: I think we should remove the `pretty_operations` computation (and all the related code in `_compute_transaction`), but in this case, the `Transaction` object should take care of generating nice update operations from the `operations` attribute.

cournape:
@johntyree can you give an ETA to get this in a state where it can be merged in master ?

johntyree:
@cournape this is today's project. It should be ready for re-review by end of day.

simplesat/transaction.py
johntyree: Well that took _forever_.

Should be all set now.

simplesat/tests/pil_pillow_upgrade_conflict.yaml
johntyree: They are now.

simplesat/transaction.py
johntyree: Added tests for `Transaction` and the `graph` module.

simplesat/transaction.py
johntyree: These have been removed entirely.

simplesat/utils/graph.py
johntyree: :+1:

cournape:
LGTM

PR #124, opened 2016-01-25
Fix requirement inconsistency

cournape:
This fixes an inconsistency when creating `Requirement` instances in the case of no version constraint. Depending on which ctor we were using, we would get either an empty set (`from_constraints`) or a set with `Any` (`_from_string`).

I also converted our usage from `set`/`frozenset` to `tuple` to keep the list of constraints, to make pretty string representation of `Requirement` more "consistent".

@johntyree

### DIFF
--- a/simplesat/constraints/multi.py
+++ b/simplesat/constraints/multi.py
@@ -30,9 +30,9 @@ class MultiConstraints(object):
 
     def __init__(self, constraints=None):
         if constraints is None:
-            self._constraints = frozenset()
+            self._constraints = tuple()
         else:
-            self._constraints = frozenset(constraints)
+            self._constraints = tuple(constraints)
 
     def matches(self, version_candidate):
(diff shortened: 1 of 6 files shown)


cournape:
Trying to figure out why travis-ci does run builds

johntyree:
`tox` passes for me with all four of our current interpreters. Not sure about travis-ci. I noticed on their status page that they had some trouble with build logs not appearing, even when builds were running. Maybe that's happening again?

cournape:
@johntyree yes, the issue was purely a travis-ci one. Looks like it is working again now.

johntyree:
LGTM

cournape:
@johntyree I realized you were right w/ the` ("*",)` vs `()` to avoid bugs BTW :)

johntyree:
Did it get you? It's sneaky. :snake:

PR #126, opened 2016-01-27
BUG: fix '<name> *' requirement parsing.

cournape:
@johntyree

### DIFF
--- a/simplesat/constraints/tests/test_requirement.py
+++ b/simplesat/constraints/tests/test_requirement.py
@@ -144,7 +144,18 @@ class TestRequirementFromString(unittest.TestCase):
         self.assertTrue(requirement.matches(V("1.9.0-1")))
         self.assertEqual(requirement, r_requirement)
         self.assertEqual(requirement, r_requirement_empty)
+        self.assertEqual(r_requirement, r_requirement_empty)
+
+        # Given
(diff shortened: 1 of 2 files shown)


simplesat/constraints/tests/test_requirement.py
```diff
@@ -144,7 +144,18 @@ def test_any(self):
         self.assertTrue(requirement.matches(V("1.9.0-1")))
         self.assertEqual(requirement, r_requirement)
         self.assertEqual(requirement, r_requirement_empty)
+        self.assertEqual(r_requirement, r_requirement_empty)
+
+        # Given
+        requirement_string = "numpy *"
```
johntyree: I actually thought this was invalid for the preamble, but I suppose this makes sense. Is the thought that someone might type this on the command line?  `jaguar install 'numpy *'` ?

cournape: they actually appear because we translate `Any()` into `*` when doing `str(requirement)`, so those come up anytime a package has a dependency in name only.

The alternative would be to remove `*` from `str(requirement)`

johntyree: Aha. Right. OK sure why not.

johntyree:
LGTM

PR #127, opened 2016-01-27
FEAT: add support for Repository.__contains__.

cournape:
A small addition to `Repository` to allow `package in repository` syntax.

@johntyree

### DIFF
--- a/simplesat/tests/test_repository.py
+++ b/simplesat/tests/test_repository.py
@@ -6,6 +6,7 @@ from okonomiyaki.versions import EnpkgVersion
 
 from simplesat.constraints import PrettyPackageStringParser
 from simplesat.errors import NoPackageFound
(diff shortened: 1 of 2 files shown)


simplesat/tests/test_repository.py
```diff
@@ -161,3 +166,31 @@ def test_find_packages(self):
...
+        dummy_with_proxy_softlink 1.0.0-1
+        nose 1.2.1-1
+        nose 1.3.0-1
+        nose 1.3.0-2\
+        """)
+        packages = self.packages_from_definition(packages_definition)
+        repository = Repository(packages)
+
+        # When/Then
+        for package in repository:
```
johntyree: It might also be good to remake a package from scratch and check that contains still works to find the original. The implementation is intended to support that, right?

johntyree:
One question, then LGTM

PR #128, opened 2016-01-28
Pass other args through to solve.py when comparing with composer

johntyree:
This just allows one to pass the other arguments to `script/solve.py` along, such as `--no-prefer-installed`.


cournape:
At, the beauty of arg management in `sh`...

LGTM

johntyree:
Isn't it beautiful? It's art.

PR #129, opened 2016-01-28
REF: split policy module into subpackage

johntyree:
This is just a refactor.


cournape:
I think some removal are missing, correct ?

johntyree:
Huh. Yeah I'm not sure what happened there. The old policy module should be removed. I'll look at it today.

johntyree:
@cournape

cournape:
I am assuming the code did not change beyond import fixes, in which case LGTM

PR #130, opened 2016-01-29
Speed up the Policy and AssignmentSet

johntyree:
This is a collection of changes I made while squeezing the blood out of the `AssignmentSet` and `UndeterminedClausePolicy`. I'm not sure how to make the hot spots in the profiler too much faster without just coming up with a smarter way to suggest packages and altogether reduce the number of calls and value checks necessary.


johntyree:
@cournape 

This actually makes the `UndeterminedClausePolicy` easier to understand in my opinion. Now it clearly lays out the order in which it checks for suggestions and does so by just searching for the maximum package version among the set of unassigned packages. What do you think?

johntyree:
For reference, this cuts `slow_bokeh_blaze.yaml` from 12s to 6.5s on my machine.

johntyree:
@cournape we should merge this so that I don't have to duckument the current version and then throw it away.

johntyree:
@cournape can we get a thumbs up on this?

cournape:
:100:

PR #131, opened 2016-01-30
 Introduce priority queue-based policy again... again

johntyree:
# This is based off of #129 and should  be merged after that one.

This is a remake of #81, which itself is a remake of #65 and #55.

This introduces the PriorityQueuePolicy again. It has good and bad qualities in comparison with the `UndeterminedClausePolicy`. It doesn't always produce the same results, but there's no reason to say that that policy is producing optimal results in the first place. They have comparable speed. On the slow scenario `slow_bokeh_blaze.yaml`, it's about 20% faster. However, when running the test suite, it's about 20% slower.

I think it has the potential to be more flexible than what we have now, but since we don't really know what the optimal way to suggest packages is I'm not sure if that's useful yet.

That all said, it's good to have a second implementation to keep things in perspective. I'll modify the tests such that they run for both policies, regardless of what we choose for a default.


johntyree:
Right now, our error messages depend on the path that the solver takes through the search space. This means that any test involving error messages is likely to fail for one policy or the other.

PR #134, opened 2016-02-13
Improved error messages on failure.

### LINKED ISSUE #132: Uninformative unsat error
When using one of the `epd_full.yaml` scenario w/ the following request:

```
request:
  - operation: "install"
    requirement: "EPD"
  - operation: "install"
    requirement: "numpy > 1.8.0"
```

I see this:

```
UNSATISFIABLE: Conflicting requirements:
Requirement: 'numpy > 1.8.0-0'
    Install command rule (+numpy-1.8.0-1 | +numpy-1.8.0-2 | +numpy-1.8.0-3 | +numpy-1.8.1-1 | +numpy-1.8.1-2 | +numpy-1.8.1-3 | +numpy-1.9.2-1)

ELAPSED : Find Requirements    : 1.224041e-03
ELAPSED : Generate Rules       : 7.586472e-01
ELAPSED : Solver Init          : 2.482009e-01
ELAPSED : SAT Solve            : 5.968580e-01
```

Which is not very helpful

johntyree:
This PR improves the `SatisifiabilityError` message significantly.

Our strategy is to dig around in the history trail of the conflicting clauses to find the root requirements, and then do a BFS to find a path of rules from one to the other. The solver's failure implies that all such paths have at least one conflict in them.

The only change in the `RulesGenerator` is that we now keep track of every rule's entire history of requirements. So if we want to install `shapely`, which depends on `numpy` which depends on `MKL`, the rule for installing `MKL` will have attached to it a list containing `numpy`'s dependence on `MKL` and `shapely`'s dependence on `numpy`.

The biggest change is in the `UNSAT` object, which now computes the path between conflicting clauses. It does this using the bog-standard queue-based BFS where all clauses that share at least one variable share an edge. In this way, we can go from clause to clause until we reach our stopping point.

For example, suppose we have a set of clauses:
```
(1, 2, -3, 4)
(-3, -4)
(2,)
(-9,)
(2, 5, 7, 8)
(-2, 9)
```

Then our path might be

```
(2,)
(-2, 9)
(-9)
```

Fixes #132 

# TODO:

- [x] Document what is going on here.

### DIFF
--- a/scripts/solve.py
+++ b/scripts/solve.py
@@ -32,7 +32,7 @@ def solve_and_print(request, remote_repositories, installed_repository,
             print(transaction)
     except SatisfiabilityError as e:
         msg = "UNSATISFIABLE: {}"
-        print(msg.format(e.unsat.to_string(pool, detailed=debug)))
+        print(msg.format(e.unsat.to_string(pool)))
         print(e.unsat._find_requirement_time.pretty(fmt), file=sys.stderr)
 
     print(solver._last_rules_time.pretty(fmt), file=sys.stderr)
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -111,6 +111,7 @@ def _connected_packages(solution, root_ids, pool):
         if name in root_names
     )
 
+    # FIXME: can use package_lit_dependency_graph() here
     def neighborfunc(pkg_id):
         """ Given a pkg id, return the pkg ids of the immediate dependencies
         that appeared in our solution. """
--- a/simplesat/rules_generator.py
+++ b/simplesat/rules_generator.py
@@ -284,9 +284,11 @@ class RulesGenerator(object):
                 if requirements is not None
                 else None)
 
-            assert len(dependency_candidates) > 0, \
-                ("No candidates found for requirement {0!r}, needed for "
-                 "dependency {1!r}".format(pkg_requirement.name, package))
+            if not dependency_candidates:
+                msg = ("No candidates found for requirement {0!r}, needed for "
+                       "dependency {1!r}")
+                raise NoPackageFound(
+                    msg.format(pkg_requirement.name, package))
 
             rule = self._create_dependency_rule(
                 package, dependency_candidates, RuleType.package_requires,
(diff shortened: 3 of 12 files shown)


johntyree:
@cournape this is ready for review

johntyree:
For reference, it looks like this right now.

```
$ python scripts/solve.py simplesat/tests/simple_numpy_installed_blocking_downgrade.yaml
UNSATISFIABLE: Conflicting requirements:
Requirements: 'numpy < 1.8-0'
    Install command rule (+numpy-1.7.1-1)
Requirements: 'numpy' <- 'numpy < 1.8-0'
    Can only install one of: (+numpy-1.8.1-1 | +numpy-1.7.1-1)
Requirements: 'numpy == 1.8.1-1' <- 'EPD_free'
    EPD_free-7.4-1 requires (+numpy-1.8.1-1)
Requirements: 'EPD_free'
    Install command rule (+EPD_free-7.4-1)
```

johntyree:
@cournape how about this?

johntyree:
@cournape I've added more comments and small explanation in the issue description. I can try to expand on this further, but if you have some questions or can point to parts that are unclear it will be more efficient.

cournape:
Ok, so this makes more sense to me now: IIUC, in the scenario "epd full", we would try to find a path between the job rule "install EPD" and "install numpy > 1.8" which includes the learnt conflicting clause(s). I am still not sure why we use a BFS for the search here ?

johntyree:
There are a few things going on. The main one is that the "learned"
clause's trail includes a bunch of clauses we really don't care about
because of the way we generate multiple clauses per rule for things like
dependencies and conflicts. So we might start out with a list of 150
clauses that are "related" in the sense that the solver looked at them
before deciding that it had found a conflict, but not related directly
enough for us to want to use them. By doing a search from the failing
requirements to one another, we find a minimal set of clauses to capture
the problem because BFS finds shortest paths.

The second one, which I'm fixing right now, is that a conflict may be
because of more than just two requirements, in which case we should make
sure to touch all of them.

2016-02-23 9:02 GMT-06:00 David Cournapeau <notifications@github.com>:

> Ok, so this makes more sense to me now: IIUC, in the scenario "epd full",
> we would try to find a path between the job rule "install EPD" and "install
> numpy > 1.8" which includes the learnt conflicting clause(s). I am still
> not sure why we use a BFS for the search here ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/enthought/sat-solver/pull/134#issuecomment-187731364>.
>



-- 
John Tyree
Scientific Software Developer
Enthought, Inc.
www.enthought.com

cournape:
So I forgot that BFS was the way to find a shortest path in unweigthed graph. That we want to find a shortest path is indeed fairly intuitive :)

cournape:
Which is to say, LGTM

johntyree:
I don't want this to grow any larger, so I'll merge it now. There are a few more improvements to make for another PR.

PR #138, opened 2016-02-23
Clean up of error message improvement code

johntyree:
This adds more comments and docstrings to the conflict handling code in UNSAT.

A major change is that the BFS now correctly returns paths that touch *all* of the end_points you give it. Before we simply stopped after the first one and got lucky because we had no tests for conflicts caused by three or more requirements together. We should catch those now and I've added a scenario with a made-up one in it.

This also fixes a non-determinism bug where python2 gives an arbitrary ordering to subclasses of `object`, so calls to `max`, `sorted` etc can differ based on the hash seed.

### DIFF
--- a/simplesat/rules_generator.py
+++ b/simplesat/rules_generator.py
@@ -63,8 +63,8 @@ class PackageRule(object):
     def __init__(self, literals, reason, requirements=None):
         self.literals = tuple(sorted(literals))
         self._reason = RuleType(reason)
-        assert isinstance(requirements, (list, type(None)))
-        self._requirements = requirements or []
+        assert isinstance(requirements, (tuple, type(None)))
+        self._requirements = requirements or ()
 
     @property
     def is_assertion(self):
@@ -176,8 +176,8 @@ class RulesGenerator(object):
             Sequence of packages that fulfill the requirement.
         reason: RuleType
             A valid PackageRule.reason value
-        requirements: list of Requirement
-            Optional requirements explaining that rule origin.
+        requirements: tuple of Requirement
+            Optional requirements explaining the rule's origin.
 
         Returns
         -------
@@ -206,8 +206,8 @@ class RulesGenerator(object):
             Package causing the conflict
         reason: RuleType
             One of PackageRule.reason
-        requirements: list of Requirement
-            Optional requirements explaining that rule origin.
+        requirements: tuple of Requirement
+            Optional requirements explaining the rule's origin.
 
         Returns
         -------
@@ -230,6 +230,8 @@ class RulesGenerator(object):
             List of packages to choose from
         reason: RuleType
             One of PackageRule.reason
+        requirements: tuple of Requirement
+            Optional requirements explaining the rule's origin.
 
         Returns
         -------
@@ -248,10 +250,14 @@ class RulesGenerator(object):
         ----------
         package: PackageInfo
             The package with a requirement
+        reason: RuleType
+            One of PackageRule.reason
+        requirements: tuple of Requirement
+            Optional requirements explaining the rule's origin.
 
         Returns
         -------
-        rule: PackageRule or None
+        rule: PackageRule
         """
         return PackageRule((-self._pool.package_id(package),), reason,
                            requirements=requirements)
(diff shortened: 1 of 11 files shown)


johntyree:
This is an improvement over what we have and should be merged if the code is OK, but it can still fail for strange or complicated errors.

Coming up with something good here is an ongoing problem. I'm tracking ideas in the wiki now. https://github.com/enthought/sat-solver/wiki/Unsatisfiability-Error-Messages

johntyree:
I've been trying some pretty gnarly things lately and this has been holding up well. I've updated the wiki to explain how it works.

Ready for review @cournape

johntyree:
We might prefer to output *all* failed paths for now. We should compare this with whatever composer is doing. It's a lot of information, but perhaps that's just an organizational problem. When we decide how to present that nicely to users we will be able to narrow it down to something digestible.

cournape:
Can you take care of seeing what composer is doing ? I tried to look into dng/hawkey, but gave up given the lack of doc

johntyree:
Composer does a pretty sexy job of this. Clearly we are losers.

For example for the scenario [`epd_full_conflict.yaml`](https://github.com/enthought/sat-solver/blob/a2a38e4152a5084e53b304aab00975c553ebc89b/simplesat/tests/epd_full_conflict.yaml):

```
request:
  - operation: "install"
    requirement: "EPD"
  - operation: "install"
    requirement: "numpy > 1.8.0"
```

We produce this, which seems partially incomplete because you could maybe try other versions of EPD or numpy. We could expand it as outlined [in the wiki](https://github.com/enthought/sat-solver/wiki/Unsatisfiability-Error-Messages#implementation-2-proposed).
```
Conflicting requirements:
      Requirements: 'EPD' <- 'numpy == 1.6.0-5'
          EPD-7.1-1 requires (+numpy-1.6.0-5)
      Requirements: 'EPD' <- 'numpy'
          Can only install one of: (+numpy-1.8.1-1 | +numpy-1.6.0-5)
      Requirements: 'numpy > 1.8.0-0'
          Install command rule (+numpy-1.8.0-1 | +numpy-1.8.0-2 | +numpy-1.8.0-3 | +numpy-1.8.1-1)
```

And here is what composers manages to say. It's not organized very coherently, but I think it has more of the pieces in place.

```
Exception:
  Problem 1
    - Conclusion: don't install EPD 7.3-2
    - EPD 7.2-1 requires numpy 1.6.1.0-patch2 -> satisfiable by numpy[1.6.1-2].
    - EPD 7.2-2 requires numpy 1.6.1.0-patch2 -> satisfiable by numpy[1.6.1-2].
    - EPD 7.3-1 requires numpy 1.6.1.0-patch2 -> satisfiable by numpy[1.6.1-2].
    - Conclusion: don't install numpy 1.6.1-2
    - Conclusion: don't install numpy 1.8.1-1
    - EPD 7.1-2 requires numpy 1.6.1.0-patch1 -> satisfiable by numpy[1.6.1-1].
    - Conclusion: don't install numpy 1.6.1-1
    - Conclusion: don't install numpy 1.8.0-3
    - EPD 7.1-1 requires numpy 1.6.0.0-patch5 -> satisfiable by numpy[1.6.0-5].
    - Conclusion: don't install numpy 1.6.0-5
    - Conclusion: don't install numpy 1.8.0-2
    - EPD 7.0-1 requires numpy 1.5.1.0-patch1 -> satisfiable by numpy[1.5.1-1].
    - EPD 7.0-2 requires numpy 1.5.1.0-patch2 -> satisfiable by numpy[1.5.1-2].
    - Can only install one of: numpy[1.8.0-1, 1.5.1-1].
    - Can only install one of: numpy[1.8.0-1, 1.5.1-2].
    - Installation request for numpy > 1.8.0.0 -> satisfiable by numpy[1.8.0-1, 1.8.0-2, 1.8.0-3, 1.8.1-1].
    - Installation request for epd * -> satisfiable by EPD[7.0-1, 7.0-2, 7.1-1, 7.1-2, 7.2-1, 7.2-2, 7.3-1, 7.3-2].
```


That said, what we have in this PR is a big improvement over what we have in master right now. Any attempt to implement idea 2 in the wiki should be a new PR.

We should also probably spend a little effort to consolidate things after we've gotten all the pieces together. Then we can just say nice things like `No version of <requirement> is compatible with <other requirement>`

cournape:
Agreed on merging that now. LGTM

PR #139, opened 2016-02-25
Use keyword arg syntax when constructing RulesGenerator

johntyree:
Wasted a bit of time before I realized that most call-sites weren't doing this and my args were ending up in the wrong order.

Someday the world will be on Python 3 and we can pretend this never happened.

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -85,7 +85,8 @@ class DependencySolver(object):
         for package in installed_repository:
             installed_map[pool.package_id(package)] = package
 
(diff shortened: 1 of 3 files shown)


cournape:
@johntyree is this change required in relation to another PR ? Because as it, since there is only 3 arguments to `RulesGenerator`, I am not sure I understand why this would change anything

johntyree:
It happened because of another PR, but I'd have done it anyway. Using keyword arguments as positional arguments has caused enough grief for me in the past that I've adopted a "don't do this" rule.

cournape:
LGTM

PR #140, opened 2016-02-26
Support ad-hoc relaxing of dependency requirements

johntyree:
This PR adds support for the notion of `allow_newer`, `allow_older`, and `allow_any` as described on this wiki page https://github.com/enthought/sat-solver/wiki/Ad-hoc-Constraints-Specification.

The general idea is that these things are specified by the user and therefore belong on the `Request` object. The `RulesGenerator` has been extended to now examine this new information on the incoming `Request` and transform the `Constraint` objects on the dependency `Requirement` objects as it creates them.

Additionally, when a transformation results in a new requirement as determined by `new == orig`, the new object will be of type `TransformedRequirement` and will have the original requirement attached to it, although it is currently private.


johntyree:
I like this in that it was pretty simple to implement and didn't require sweeping changes. However, I worry because there were a few cases where I was forgetting to pass along these extra things and it caused subtle bugs.

For example, if you are trying to sort out the order in which things should be installed and you forget to apply the transformations to the dependencies, your install plan might not make sense. It would be more robust if the entire repositories were swapped out with new ones that had the transformed metadata. Then you couldn't forget to use them, but I'm not sure how it affects our "packages are immutable" assumption. Also, that's a *lot* of packages. I can't imagine it would perform well.

At any rate, this is ready for some review. We can talk about persisting these rules in another PR.

cournape:
How complicated would it be to change the strategy later to swap repositories ? I would actually hope that thanks to immutable packages, it would not be complicated to create a new repository where only a few packages may be changed

simplesat/constraints/tests/test_requirement.py
```diff
@@ -265,3 +271,173 @@ def test_repr(self):
...
+class TestRequirementTransformation(unittest.TestCase):
+
+    name = "MKL"
+    not_name = "shapely"
+    never_name = "basemap"
+
+    def run_transformations(self, constraints, transform, allow):
+
+        self.assertSingleMulti(
+            constraints, self.name, transform_install_requires, allow)
```
cournape: shouldn't this be `transform` ?

simplesat/constraints/tests/test_requirement.py
```diff
@@ -265,3 +271,173 @@ def test_repr(self):
...
         # Then
         self.assertMultiLineEqual(repr(requirement), r_repr)
+
+
+class TestRequirementTransformation(unittest.TestCase):
+
+    name = "MKL"
+    not_name = "shapely"
+    never_name = "basemap"
+
```
cournape: I am not sure I understand the variable names here ?

simplesat/constraints/tests/test_requirement.py
```diff
@@ -265,3 +271,173 @@ def test_repr(self):
...
+            # Then
+            self.assertEqual(expected, result)
+
+    def _add_name(self, name, pairs):
+        return tuple(
+            (name + ' ' + before, name + ' ' + after)
+            for before, after in pairs)
+
+    def _nub(self, sequence):
```
cournape: what does `nub` stand for ?

simplesat/constraints/tests/test_requirement.py
```diff
@@ -265,3 +271,173 @@ def test_repr(self):
...
         self.assertMultiLineEqual(repr(requirement), r_repr)
+
+
+class TestRequirementTransformation(unittest.TestCase):
+
+    name = "MKL"
+    not_name = "shapely"
+    never_name = "basemap"
+
+    def run_transformations(self, constraints, transform, allow):
```
cournape: the name of that function (and the `assert*`) made it a bit difficult for me to follow.

Normally, the convention for `assert*` methods is to have the signature as `assertFoo(result, expected_result, ...)`. I think it would help changing the signature to that convention.

Also, it is not obvious what `assertSingleMulti` would do without looking at the code, same for `assertTransformation`. Can we get more explicit names there as well ?

simplesat/tests/adhoc_allow_newer_older_any.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - operation: "install"
+    requirement: "B ^= 1.0.0"
+  - operation: "install"
+    requirement: "E"
+
+transaction:
+  - kind: "remove"
+    package: "A 0.0.1-1"
+  - kind: "install"
+    package: "A 2.0.0-1"
```
cournape: isn't that an implementation detail ? Strictly speaking, one could get another solution which includes `A ^= 1.0.0`, right ? Does it make sense to add in `request` an `install A ^= 2.0.0` operation ?

cournape:
A few naming comments, as I found understanding the tests a bit difficult for what it actually does, but otherwise LGTM !

simplesat/constraints/tests/test_requirement.py
johntyree: Yes. :disappointed:

johntyree:
> How complicated would it be to change the strategy later to swap repositories ? I would actually hope that thanks to immutable packages, it would not be complicated to create a new repository where only a few packages may be changed.

This could work too, we just need to decide how to make sure that no one uses the "old" untransformed repository after the sat-solver has computed a new one.

johntyree:
Although, we still have that problem now because everyone has to remember to apply the transformation themselves.

simplesat/tests/adhoc_allow_newer_older_any.yaml
johntyree: I don't think anything about the set of packages that are chosen should be an implementation detail. Right now we have an explicit preference for already installed packages followed by the newest possible packages that satisfy the constraints.

johntyree:
@cournape I managed to sort it out by having the pool be aware of the request and recomputing packages as necessary. This is nice because now you can just pass the pool around and everything will behave without remembering to do anything else. The trade-off is that the pool contains packages with *unparsed* constraints, so we can't pass around the transformed requirement objects themselves. Instead what I've done is attached a reference to a `parent_package`, which is non-null only when the package has been derived from another because of constraints.

If you think this is nicer, we can stick with it. Otherwise I'll reset back to https://github.com/enthought/sat-solver/commit/80c7d39e16cb777786e57d416721c54644e302c7.

johntyree:
@cournape I think this clears it up pretty well. Ready for re-review.

johntyree:
## Note to self, #147 is against *this*, not `master`. Merge that here before merging this there.

johntyree:
I'll update this to only attach the `ConstraintModifiers` and not the entire `Request` object.

johntyree:
@cournape  I think this is OK now.

simplesat/pool.py
```diff
@@ -1,26 +1,59 @@
...
+        self._original_packages = []
+        # FIXME Mar-9-2016: temporarily changing these names to catch places
+        # that were using the private API. If it has been awhile and you feel
+        # like everything is ok, you can remove these trailing underscores.
+        self._package_to_id_ = {}
+        self._id_to_package_ = {}
+        self._packages_by_name_ = DefaultOrderedDict(list)
+
+    def modifiers():
+        def fget(self):
```
cournape: can't we use the "normal" decorator syntax (`@`) ?

johntyree:
I take that back. There are actually quite a few places internally where we assume that whatever goes into the pool is what will come out again. Maybe it would be better to push the `ConstraintModifier` even farther down and onto the `Repository` objects themselves? I'll think about it tomorrow.

johntyree:
OK. I think I will try a hybrid of the two approaches. We'll attach the `ConstraintModifiers` to the `Pool` in order to prevent accidentally forgetting to use them, but we *won't* output transformed packages. We will only transform `Requirement` objects as they are passed to `.what_provides()`.

We can cache these as well to avoid re-computing the transformation.

johntyree:
This PR now introduces two subclasses of `Requirement`. They add no new features, but are there to maintain some modicum of safety, now that the `Pool` does the transformation without knowing where the requirement came from.

The requirement transformation code will only accept `InstallRequirement` objects to prevent accidentally transforming `Requirement` objects made from the `.constraints` information. We may decide that there are reasonable transformations to be done there, but that's a choice for later and will fit into this just fine. Since it's essentially just a rename for most code, this is a pretty safe refactor.

simplesat/constraints/constraint_modifiers.py
```diff
@@ -0,0 +1,151 @@
...
+import six
+
+from okonomiyaki.versions import EnpkgVersion
+from simplesat.constraints.kinds import (
+    Any, EnpkgUpstreamMatch, Equal, Not, GEQ, GT, LEQ, LT,
+)
+from simplesat.constraints.requirement import InstallRequirement
+
+
+MAX_BUILD = 999999999  # Nine nines... I guess
```
johntyree: See #151

simplesat/constraints/constraint_modifiers.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+
+class _TransformRequirement(object):
+
+    @staticmethod
+    def __call__(requirement, **kw):
+        if isinstance(requirement, InstallRequirement):
+            return _transform_install_requirement(requirement, **kw)
+        elif requirement.has_any_version_constraint:
```
johntyree: This will allow non-`InstallRequirement` objects to pass through, as long as the transformation is a no-op.

Should it be a `TypeError` instead?

simplesat/constraints/constraint_modifiers.py
```diff
@@ -0,0 +1,151 @@
...
+            new_r = _transform_install_requirement(requirement, **kw)
+            msg = "Only identity transformations are defined for {}"
+            class_name = requirement.__class__.__name__
+            if new_r is not requirement:
+                raise NotImplementedError(msg.format(class_name))
```
johntyree: Should it be a `TypeError` instead?

simplesat/rules_generator.py
```diff
@@ -391,6 +393,13 @@ def _add_install_job_rules(self, job):
                 packages, RuleType.job_install,
                 requirements=(job.requirement,))
             self._add_rule(rule, "job")
+        else:
+            requirement = self._pool.transform_requirement(job.requirement)
+            msg = str(requirement)
+            if requirement is not job.requirement:
```
johntyree: Maybe better to use `==` here and not rely on identity...

simplesat/test_utils.py
```diff
@@ -124,7 +125,8 @@ def from_yaml(cls, file_or_filename):
             if kind == 'update_all':
                 update_all = True
                 continue
-            requirement = Requirement._from_string(s_request["requirement"])
+            requirement = InstallRequirement._from_string(
```
johntyree: Should be a normal `Requirement`. User job requirements don't get transformed.

johntyree:
@cournape OK ready to try again, again.

simplesat/pool.py
```diff
@@ -29,15 +34,15 @@ def add_repository(self, repository):
...
+            self._id_to_package_[current_id] = package
+            self._package_to_id_[package] = current_id
+            self._packages_by_name_[package.name].append(package)
 
-            self._id_to_package[current_id] = package
-            self._package_to_id[package] = current_id
-            self._packages_by_name[package.name].append(package)
-
-    def what_provides(self, requirement):
+    def what_provides(self, requirement, transform=True):
```
cournape: can we rename `transform` into e.g. `use_modifiers` ? This way it is clearer what transform refers to IMO

cournape:
One suggestion, LGTM otherwise

johntyree:
Yeah I thought about this as well, I'll come up with something.

cournape:
@johntyree feel free to merge this as soon as you make the change.

johntyree:
:dancer: :tada:

PR #141, opened 2016-02-26
Support ad-hoc tightening of constraints

johntyree:
# This includes #140. Look there first.

This PR adds support for arbitrary constraints as "jobs". It looks like this in the scenario files.

```yaml
request:
  - operation: "constrain"
    requirement: "B ^= 1.2.0"
  - operation: "install"
    requirement: "E"
  - operation: "constrain"
    requirement: "F > 2.0"
```

It was a little tricky to implement because our clauses are not very expressive. Thus, if you want to create a constraint that says "If MKL is being installed, install a version >= 11," you either need to know whether or not MKL is being installed, which means changing the way the `RulesGenerator` works to include a preprocessing step that sees all the rules before they are run through the `pool` and turned into package literals... *OR* you make a pile of rules that together say "Do not install any version of MKL < 11."

For example, if you had some MKL packages:
```
packages:
  - MKL 10.2-1
  - MKL 10.3-1
  - MKL 11.0-1
  - MKL 11.1-1
```
and you want to constrain any solution to use `MKL >= 11`, you can create two rules that say `NOT MKL 10.3-1` and `NOT MKL 10.2-1` instead. This is how it's currently done.

I was somewhat worried about creating more rules and clauses, but since they are already unit, they should immediately force assignments to their literals and not get caught up in the churn of solving.


simplesat/rules_generator.py
```diff
@@ -399,6 +406,49 @@ def key(package):
...
+
+        For example the install job for "MKL > 10.0" would translate to "Must
+        install an MKL version greater than 10.0.0-0," whereas a constrain job
+        would be read as "Must not install an MKL version less than or equal to
+        10.0.0-0."
+
+        """
+        # XXX: an alternative approach might be to actually negate the
+        # constraint, but that requires more logic and seems more likely to
+        # need special rules for edge cases.
```
johntyree: Actually maybe it wouldn't be much harder? `pool.what_provides(R(job.requirement).negated())` isn't so crazy. Would that work in all cases?

cournape:
Can we discuss this offline ? I have to confess I don't quite understand the problem we are solving here

johntyree:
I'll revisit this after addressing the problems with #140

johntyree:
:boom:

I have another PR for this rebased on top of the improved #140, see https://github.com/enthought/sat-solver/pull/147.

PR #145, opened 2016-03-03
BUG: don't trim long php exception error messages

johntyree:
When they are very long, php doesn't display the whole thing, for whatever reason.

Here I'm just printing it out. It's usually going to be a description of a conflict and might be lengthy.


cournape:
Well, whatever works when php is concerned ! :+1:

PR #146, opened 2016-03-03
ENH: be stricter about distribution name and version parsing

johntyree:
This tries hard to be more strict about what we accept as a valid distribution name.

I've tested on all the distribution names in the index and was able to round trip them all through a Requirement object successfully.

### DIFF
--- a/simplesat/constraints/parser.py
+++ b/simplesat/constraints/parser.py
@@ -144,7 +144,7 @@ def _tokenize(scanner, requirement_string):
         for part in requirement_string.split(","):
             scanned, remaining = scanner.scan(part.strip())
             if len(remaining) > 0:
-                msg = "Invalid requirement string: {0!r}"
+                msg = "{0!r}"
                 for tok in scanned:
                     if isinstance(tok, DistributionNameToken):
                         msg += "(distribution name: {0!r})".format(tok.value)
@@ -177,8 +177,7 @@ class _RawConstraintsParser(object):
                 assert isinstance(requirement_block[0], AnyToken)
                 return Any()
             else:
-                msg = ("Invalid requirement string: {0!r}".
-                       format(requirement_string))
+                msg = "{0!r}".format(requirement_string)
                 raise InvalidConstraint(msg)
 
         constraints = []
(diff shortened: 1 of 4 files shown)


simplesat/constraints/parser.py
```diff
@@ -10,16 +10,20 @@
 from simplesat.errors import SolverException
 
 
-_VERSION_R = "[^=><!,\s;^][^,\s;]+"
-_EQUAL_R = "=="
-_GEQ_R = ">="
+# NOTE: re.Scanner seems to ignore (?i), so we have to write case-insensitivity
+# into the regex manually. We must also permit a leading underscore because of
+# names like `_distribute_remove`
```
cournape: Can we reference the PEP defining this regex, and mention we remove `-` as this breaks too many things otherwise ?

johntyree:
@cournape How's this?

johntyree:
What exactly did we want to do here? Put this into okonomiyaki?

cournape:
We can put distribution name checking in okonomiyaki later. Right now, we just want to make sure we don't reject anything valid, and get better error messages for invalid requirements. Did you have a chance to validate the regex on the table I put on slack yesterday ?

johntyree:
Sigh... `pypol_` is a package name apparently. If we fix this it will almost be no different from our regex for versions. :disappointed:

```python
_DISTRIBUTION_R = r"([a-zA-Z0-9_][a-zA-Z0-9._]*[a-zA-Z0-9]|[a-zA-Z0-9])"
_VERSION_R = r"(\d\w*(?:\.\w+)*(?:-\w+)?)"
```

johntyree:
@cournape what about this?

cournape:
Can we add more tests regarding for error handling of invalid requirement string ? Or do we want to do that in another PR ?

johntyree:
Do you mean ensuring that we fail on bad strings? I can construct artificial strings that don't match the regex, but is that useful?

cournape:
IIRC, what started the discussion was failing to parse things like "numpy<1.9", "ipython > 4", etc... and give decent error messages ?

johntyree:
@cournape so now we test for several different badly formatted package names and versions as well.

Decent error messages happen higher up the stack. I'll check that we do those correctly next.

johntyree:
@cournape OK I think we're there. How's this?

cournape:
LGTM

PR #147, opened 2016-03-07
 Support ad-hoc tightening of constraints 2

johntyree:
# This PR is against #140, *NOT* `master`. This is for comparison.

This is a rebase of #141 on top of the improved #140.

I'm still researching the various uses of this, but as you can see, our current architecture gives it to us very cheaply, regardless.

One use for this from an IT perspective might be simply to exercise control over the package set deployed site-wide. An administrator, independent of any particular application, might be interested in preventing a particular package version from being installed. For example, pinning `FooBar` to `>= 0.18, < 0.19` until they can certify the 0.19 series.

My original idea was to use this feature in conjunction with some kind of list of blessed versions to help in recreating an environment, for example via a Cargo-style lock file.

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -62,20 +62,25 @@ class DependencySolver(object):
 
         for job in request.jobs:
             assert job.kind in (
-                JobType.install, JobType.remove, JobType.update
+                JobType.install, JobType.remove, JobType.update,
+                JobType.constrain,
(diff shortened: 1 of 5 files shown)


johntyree:
This is on hold until #140 is in use and we have a better understanding of use cases.

johntyree:
I will close this for now. I suspect we'll want to revive it when we get to supporting clients that want to use cross platform lockfiles, but we'll cross that river when we get there :camel: :camel: :camel:

PR #148, opened 2016-03-07
MAINT: always attach requirement to NoPackageError

johntyree:
That's all we do here.

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -69,7 +69,7 @@ class DependencySolver(object):
 
             providers = tuple(pool.what_provides(requirement))
             if len(providers) == 0:
-                raise NoPackageFound(str(requirement), requirement)
+                raise NoPackageFound(requirement, str(requirement))
 
             if job.kind == JobType.update:
                 # An update request *must* install the latest package version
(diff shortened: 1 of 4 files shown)


cournape:
Can I ask to swap the argument order ? I think we always want a requirement, but we could use a default message string

cournape:
@johntyree does the comment make sense ? It would be good to get this in master and update jaguar with it

johntyree:
Yes. Just haven't gotten to it. Was doing docs instead. I'll try to wrap up
these PRs with very little work left on them today

johntyree:
What kind of default message? I was thinking the implementation would look something like
```python
class NoPackageFound(SolverException):
    def __init__(self, requirement, *a, **kw):
        super(NoPackageFound, self).__init__(*a, **kw)
        self.requirement = requirement
```
Is that sufficient?

johntyree:
Or possibly:
```python
class NoPackageFound(SolverException):
    def __init__(self, requirement, *a, **kw):
        super(NoPackageFound, self).__init__(*a, **kw)
        self.requirement = requirement
        self.args = self.args or (str(requirement),)
```

cournape:
LGTM

PR #152, opened 2016-03-14
DOCS DOCS DUCS DUCKS DUCKS DUCKS DUCKS

johntyree:
Each more beautiful than the last.


johntyree:
These aren't finished, but they describe some of the highest level relationships.

There isn't much description about the operational behavior yet, i.e. how the `UndeterminedClausePolicy` actually works, how requirements get translated into rules, etc.

docs/source/glossary.rst
```diff
@@ -0,0 +1,29 @@
+Glossary
+========
+
+  Repository
```
cournape: Requirement is missing

cournape:
wait, are goose ducks ?

johntyree:
Unfortunately, geese are not ducks.

docs/source/glossary.rst
johntyree: Yes, this glossary is still very incomplete.

There are also no cross references or anything else. I need to figure out how ReST and sphinx work.

johntyree:
Check that this works on readthedocs.

johntyree:
Also add `Requirement` to the glossary.

johntyree:
I've added the `Requirement` to the glossary and this should build correctly on readthedocs when we merge.

Merging now!

PR #154, opened 2016-03-14
REF: only RuleType knows which members are jobs

johntyree:
Small refactor.

The `RuleType` is where we should track which types come directly from the user.


johntyree:
Maybe this makes more sense as a property rather than a function.

johntyree:
@cournape works nicely thanks.

cournape:
LGTM

PR #155, opened 2016-03-14
ENH: track package requirements through the work_queue

johntyree:
This fixes a long-standing annoyance with the error reporting. It also enables us to rely on the requirement chain length, which makes it easier to figure out if a particular set of requirements came from a user (because there's only one) or if it was derived by following the dependency graph (because there's more than one).


cournape:
:+1:

Issue #144, opened 2016-03-02
missing depedencies throw assertion errors instead of dependency solving error


itziakos:
for example `legacy`

```python
PS C:\Users\itziakos\Projects\haas> jaguar --no-debug install ipython-3.0.0-1
Fetching package repositories indices.. done
One of the requested package has broken dependencies
(Dependency solving error: Could not resolve "terminado 0.5" required by "RemotePackageMetadata('ipython-3.0.0-1', key='ipython-3.0.0-1.egg', available=True, product='free', repository_info='BroodRepository(<https://packages.enthought.com>, <enthought/free>)')")
```
while `sat_solver`

```python
PS C:\Users\itziakos\Projects\haas> jaguar --no-debug install ipython-3.0.0-1
Fetching package repositories indices.. done
Unexpected error
AssertionError("No candidates found for requirement 'terminado', needed for dependency RepositoryPackageMetadata('ipython-3.0.0-1', repo=Repository(<enthought/free>))",)
```

cournape:
We need to do 2 things here:

1. Better error handling in the SAT solver (that should not be too complicated)
2. Figure out what to do about broken dependencies when they are not explicitly required. In your case, we should fail with a message clear to the `legacy` case, but if all you want is one version of ipython, we should maybe just ignore packages with broken dependencies.

PR #156, opened 2016-03-14
Enh/missing constraints

### LINKED ISSUE #149: What to do when no candidate is found in the rules generator
At least the 2 following cases are relevant:

1. Conflicts that cannot be resolved. Should those then be ignored ?
2. Dependencies missing in some requirements, e.g. package `A 1.0.0-1` depends on `B ^= 1.0.0`, `A 1.0.0-2` does not depend on B. For some reason, no candidate for `B ^= 1.0.0` can be found. It should be possible to install `A 1.0.0-2` (ignoring missing deps for `A 1.0.0-1`), or raise some kind of broken dependency exception if  `A 1.0.0-1` is requested.

johntyree:
This allows us to optionally ignore packages with missing dependencies or conflicts if they aren't tied directly to a job from the user.

If strict is `True`:
   - When a user specified job is broken, throw an exception.
   - When a dependency or conflict is broken log a WARNING.

If strict is `False`:
   - When a dependency or conflict is broken log an INFO.

Closes #149, #144


johntyree:
Still need to check this on windows. I don't know of any broken packages of ours on Linux.

simplesat/rules_generator.py
```diff
@@ -296,20 +310,42 @@ def _add_install_requires_rules(self, package, work_queue, requirements):
...
+                    # job requirement. Unfortunately, we don't track that
+                    # explicitly because we push all of the work through a
+                    # queue. As a proxy, we can examine the associated
+                    # requirements directly.
+                    if len(requirements) == 1:
+                        raise MissingInstallRequires(pkg_requirement, msg)
+                    else:
+                        logger.warning(msg)
+                else:
+                    logger.info(msg)
```
johntyree: Missing a `return` statement.

johntyree: [Nevermind](https://github.com/enthought/sat-solver/pull/156/files#diff-27965fdafe1fac1e643870b7f741db60R337), it's down below. We have to make our rule preventing installation first.

johntyree:
Turns out `MiniSat` was not handling conflict discovery correctly during initialization. It never mattered before because we don't typically have totally broken package metadata. Now that we explicitly ban a package with broken metadata, we can end up with conflicts before we're even done reading all the rules. This fixes that.

johntyree:
This should fix #144 entirely. I've confirmed that on master, it crashes or in some unfortunate cases, doesn't crash *or* solve the problem, while on this branch it does.

There are also tests that replicate the problem and verify that it is caught correctly now.

simplesat/rules_generator.py
```diff
@@ -296,20 +310,43 @@ def _add_install_requires_rules(self, package, work_queue, requirements):
...
+                req_str = str(pkg_requirement)
+                msg = ("Blocking package {0!s}: no candidates found for"
+                       " dependency {1!r}").format(pkg_str, req_str)
+                if self.strict:
+                    # We only raise an exception if this comes directly from a
+                    # job requirement. Unfortunately, we don't track that
+                    # explicitly because we push all of the work through a
+                    # queue. As a proxy, we can examine the associated
+                    # requirements directly.
+                    if len(requirements) == 1:
```
cournape: I don't understand the logic behind testing for `len(requirements) == 1` ?

cournape:
The code LGTM, but I have one question regarding the logic

johntyree:
I guess my long comment there didn't work, :disappointed:

The behavior differs depending on whether the broken package is directly
from a job requirement or just from a transitive dependency. Unfortunately,
the only way to know without a large refactor is to count the number of
requirements in the stack.

If there's only one it's a job, Otherwise it isn't.

johntyree:
I made the comment a little more explicit. I'll merge after Travis gives the nod.

cournape:
I understood the comment, but did not get `len(requirements) == 1` implied the corresponding requirement was a job requirement.

cournape:
and I am not sure that I do yet :) I understand that for a job requirement, `len(requirements) == 1`, but not the opposite relationship

johntyree:
Forcefully installed broken package discussion can move to https://github.com/enthought/sat-solver/issues/158.

Merging this.

PR #157, opened 2016-03-14
ENH: add api for checking requirements against a repo

### LINKED ISSUE #137: Add API to sat solver to check whether a given set of requirements is consistent or not
This is needed to make the PoC for requirements support in jaguar. Ideally, we could detect which requirements are not consistent for a given set of requirements.

johntyree:
How does `DependencySolver.requirements_are_satisfiable(repositories, requirements)` look?

Closes #137


simplesat/dependency_solver.py
```diff
@@ -12,6 +13,22 @@
 
 
 class DependencySolver(object):
+
+    @classmethod
+    def requirements_are_satisfiable(cls, repositories, requirements):
```
johntyree: Let's make this a standalone function instead.

johntyree: Done.

johntyree:
@cournape if this meets whatever need you had in mind, it's ready to go.

cournape:
Ah, so I think this is indeed not what we need. The rationale for this feature was detecting inconsistent `requirements.txt` files. `requirements.txt` files are expected to be complete (that is describe a full solution to the solver), so I think we need a bit more than what this does.

For example, the following would be an invalid `requirements.txt` file:

```
numpy 1.9.2-3
scipy 0.17.0-2
```

because the MKL is missing.

johntyree:
OH. That's completely different, yes.

I think both will be useful. I'll leave this and add another for that case.

cournape:
I agree both are useful. I should have been clearer on the usecase, sorry about that

cournape:
This is also why I suggested the API to have the `installed_repository` specified separately, and then forgot about it !

So what we really need is:

1. something to check that a given (installed) repository is consistent: the idea is to check that we are starting from a consistent state.
2. something to validate that a given set of requirements is a valid, full solution
3. something that checks that given an installed repository validated by 1, we can apply a set of requirements validated by 2. The idea here is to enable applying a requirements file into an existing environment with already installed packages. I am not sure that can always be done, though

johntyree:
OK. So, ignoring the optional `modifiers`, how about this?

> Something to check that a given (installed) repository is consistent: the idea is to check that we are starting from a consistent state:
```
repository_is_consistent(repository)
```

> Something to validate that a given set of requirements is a valid, full solution.
```
requirements_are_complete(repositories, requirements)
```

> Something that checks that given an installed repository validated by 1, we can apply a set of requirements validated by 2. 
```
requirements = requirements_from_repository(installed_repository)
if requirements_are_complete(new_requirements):
    requirements += new_requirements
else:
    return False
return requirements_are_complete(repositories, requirements)
```

Do we want a function specifically for that last part? Some kind of `can_add_legacy_requirements(installed_repository, repositories, requirements)` ?

cournape:
We can do without a function for the last part I guess. The API you propose look sufficient.

johntyree:
Is that a LGTM?

cournape:
I have not looked at the changes yet, sorry. Looking at it now

cournape:
So looking at this now, I see something strange:

```
from jaguar.core import Settings
from jaguar.network import Session

from jaguar.core.packages_manager import packages_repositories_factory
from jaguar.core.requirements import parse_requirements_file
from jaguar.core._sat_solver import (
    _convert_enstaller_repository_to_solver_format
)

from simplesat import Requirement
from simplesat.dependency_solver import (
    requirements_are_satisfiable, requirements_are_complete
)


requirements = [
    Requirement._from_string("{0} == {1}".format(name, version))
    for name, version in parse_requirements_file("requirements-win32.txt")
]

settings = Settings.from_filename("/home/vagrant/.jaguar.yaml")
session = Session.authenticated_from_settings(settings)

with session:
    repositories, _unavailable = packages_repositories_factory(session, settings.repositories)
    repositories = (
        _convert_enstaller_repository_to_solver_format(repository)
        for repository in repositories
    )

    print("verifying...")
    print(requirements_are_complete(repositories, requirements))
    print("One more time")
    print(requirements_are_complete(repositories, requirements))
```

This fails the second time w/ the following exception:

```
  File "verify_requirements.py", line 34, in <module>
    print(requirements_are_complete(repositories, requirements))
  File "/home/vagrant/src/sat-solver/simplesat/dependency_solver.py", line 58, in requirements_are_complete
    [repo], requirements, modifiers=modifiers)
  File "/home/vagrant/src/sat-solver/simplesat/dependency_solver.py", line 73, in requirements_are_satisfiable
    DependencySolver(pool, repositories, []).solve(request)
  File "/home/vagrant/src/sat-solver/simplesat/dependency_solver.py", line 106, in solve
    request
  File "/home/vagrant/src/sat-solver/simplesat/dependency_solver.py", line 143, in _create_rules_and_initialize_policy
    raise NoPackageFound(requirement, str(requirement))
simplesat.errors.NoPackageFound: mkl == 10.3-2

```

johntyree:
Hmmm yes. Looks bad. Can you share the requirements file that causes this?

cournape:
```
mkl 10.3-2
numpy 1.9.2-2
libgfortran 3.0.0-2
```

cournape:
(On Linux)

johntyree:
Well that's underwhelming. I'll look at this today.

johntyree:
Something is wrong with iteration over repositories. This already fails:

```
    repositories, _unavailable = packages_repositories_factory(session, settings.repositories)
    repositories = (
        _convert_enstaller_repository_to_solver_format(repository)
        for repository in repositories
    )

    pkgs = [list(repository) for repository in repositories]
    pkgs2 = [list(repository) for repository in repositories]

    assert pkgs == pkgs2
```

johntyree:
Oh... You used a generator expression to convert the repositories :smile: 

Cleverness loses again.

Try this

```
    repositories = tuple(
        _convert_enstaller_repository_to_solver_format(repository)
        for repository in repositories
    )
```

cournape:
:( Well done David

johntyree:
:beers:

johntyree:
Does this do what we need now?

cournape:
I think so. The next step would be to get some information about what's missing/conflicting in case of errors, but that can come in another PR

johntyree:
Well right now I'm throwing that away because you asked for `f : stuff -> bool`. We can expose it pretty easily.

What about `f : stuff -> {'ok': bool, data: UNSAT}` ?

cournape:
I have not looked at it, but I suspect `UNSAT` would be too generic to be useful. I would hope we could do much better here with not too much effort.

johntyree:
UNSAT has a lot of information attached to it. It's not the exception type, just a data type characterizing unsatisfiable problems. From that object we can construct our typical error string or look at anything else we might care about. It would be useful to know what it is we are trying to do.

cournape:
Sure. So thinking about the common errors in requirements file, I can think of:

1. missing dependencies (e.g. numpy depends on libgfortran which is not in the requirements list)
2. conflicting dependencies (e.g. `numpy 1.9.2-2` depends on `MKL 10.3-2` but `MKL 11.1.4` is installed)

Both seem "easy" to detect without even going through `UNSAT` (but I could of course be wrong).

Another error would be actual conflict (e.g. 2 versions of the same package, etc...). but 1. and 2. may be worth special casing if it helps giving better error message.

Does that make sense ?

johntyree:
Currently, the way we handle errors would give us something like this for case 1.
    ```
    Requirements: 'numpy == 1.9.1-1' <- 'MKL == 10.3-1'
        +numpy-1.9.1-1 was ignored because it depends on missing packages
    Requirements: 'numpy == 1.9.1-1'
        Install command rule (+numpy-1.9.1-1)
    ```

For case 2 it's not clear to me exactly what it is we're comparing.
> conflicting dependencies (e.g. numpy 1.9.2-2 depends on MKL 10.3-2 but MKL 11.1.4 is installed)

Does this mean we're just trying to figure out if it's possible to satisfy new requirements without uninstalling anything? If so, then why aren't we just feeding all the requirements into the solver together?

PR #159, opened 2016-03-16
Show log report based on --debug

johntyree:
Show the log report if `debug` is nonzero, show it verbosely if `debug > 1`.


cournape:
:+1:

PR #160, opened 2016-03-19
DOCS DOCS DUCS DUCKS DUCKS DUCKS DUCKS 2

johntyree:
I pretty much just read through the docs for Requests and started to do something similar. It is of course not done yet, I'll just keep making PRs as I get more content written.

We're still waiting on #130 before I start writing about how those are implemented.

### DIFF
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -1,13 +1,19 @@
+.. _architecture:
+
 Architecture
 ============
 
 Simplesat is modeled after the Composer library from PHP.
 
 For a good overview of the public API of the entire system, you should look at
-``simplesat.test_utils.Scenario``, upon which all of our functional testing is
-based. The ``Scenario`` class shows how to build ``PackageMetadata`` instances
-from strings, use them to create a ``Repository``, ``Pool`` and ``Request`` and
-pass them to a ``DependencySolver`` for resolution.
+the :class:`Scenario <simplesat.test_utils.Scenario>`, upon which all of our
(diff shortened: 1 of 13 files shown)


johntyree:
New docs look like this: https://sat-solver.readthedocs.org/en/develop/api.html

docs/source/architecture.rst
```diff
@@ -1,15 +1,21 @@
+.. _architecture:
+
 Architecture
 ============
 
 Simplesat is modeled after the Composer library from PHP.
```
cournape: I think it may be more accurate to say the API is modeled after composer, since the solver itself is not really related to composer.

simplesat/dependency_solver.py
```diff
@@ -37,10 +61,21 @@ def repository_from_requirements(repositories, requirements, modifiers=None):
 
 
 def repository_is_consistent(repository, modifiers=None):
-    """ Return True if every package in the repository can be installed
-    simultaneously, otherwise False.
-
-    If `modifiers` are not None, use them when resolving requirements.
+    """
+    Return True if all packages in a repository can be installed together.
+
```
cournape: We should mention this only really makes sense for repositories representing installed packages (e.g. no more than 1 version of any package in it).

cournape:
LGTM

simplesat/dependency_solver.py
johntyree: :+1:

PR #165, opened 2016-03-21
REPO: build docs on travis

johntyree:
RTD doesn't seem to tell anyone if it fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ python:
   - "3.5"
   - "pypy"
 
+addons:
+  apt:
+    packages:
+      - graphviz
+
 install:
   - pip install -U pip
   - pip install -U setuptools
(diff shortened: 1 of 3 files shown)


johntyree:
:dancer: @cournape

cournape:
Can we have this as a separate job instead of building docs for every job ?

johntyree:
We could, but why should we? I would think "everything works on every platform" is a reasonable default to want to have and building the docs adds about four seconds to the build time.

At any rate, it isn't possible to have doc-updating PRs only run the doc job anyway, right? So everything gets run no matter what you do.

cournape:
Is it really only 4 seconds including the time to install sphinx and co ? If so, then I agree is it not worth it yet.

The main point of having a separate job is that it can run in // and not serially.

johntyree:
It looks like it adds about 10% to the build time, but there's so much variability depending on Travis itself that it's hard to tell. Right now builds take anywhere from 1min to 3min, with or without this.

cournape:
Just to make it clear: it is indeed not worth arguing about if it adds a few sec :)

:+1:

PR #166, opened 2016-03-21
Fix full package name parsing

johntyree:
Our pretty package parser was getting upset at packages that hyphenate between the distribution name and the version.

This fixes that.

### DIFF
--- a/simplesat/constraints/package_parser.py
+++ b/simplesat/constraints/package_parser.py
@@ -7,12 +7,13 @@ from .parser import _DISTRIBUTION_R, _VERSION_R, _WS_R
 
 _WS_RS = _WS_R
 _MAYBE_WS_RS = " *"
+_SEP_RS = r"(?:(?:{})|(?:{}))".format(_WS_RS, '-')
 _DISTRIBUTION_RS = "(?P<distribution>{})".format(_DISTRIBUTION_R)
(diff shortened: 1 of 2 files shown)


johntyree:
Of course... it won't pass tests if we specifically test that it doesn't allow hyphenation.

@cournape how do we want to handle this? It seems to me that `numpy-1.9.2-3` is just as valid as `numpy 1.9.2-3`.

cournape:
@johntyree I don't understand your comment given the failure. I guess that strictly speaking, `numpy1.8.1-1` is valid requirement, with the distribution name being `numpy1.8.1` and the version `1`.

johntyree:
What do you mean? The question is whether or not we want to accept a hyphen between the name and version or not.i think right now we have a test that makes sure we don't. I think we should allow it.

cournape:
I mean I don't see how that's related to the test failure.

I don't think we should allow `-` between name and version at the sat solver level, because `-` are used for other purposes, and it is hard to parse well when you go beying the model `name-upstream_version-build`. The only reason to support it is for backward compatibility, and that can be done at the jaguar level.

johntyree:
Ah OK. I've been working under the assumption that we *are* allowing that.

We really need to consolidate our package parsing then. I think there are several different places where we parse string input and they do not all have the same semantics.

I'll close this then.

PR #167, opened 2016-03-21
REF: work with package iterables, not repositories

johntyree:
None of these functions care about repositories. They are easier to use and understand if we just pass around the packages themselves.


cournape:
LGTM

PR #169, opened 2016-03-22
MAINT: add enum34 to the normal requirements file

johntyree:
@noraderam

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 setuptools>=18.4
 PyYAML
+enum34>=1.1.2


noraderam:
:+1:

cournape:
Can we add the missing runtime requirements to `setup.py` ?

johntyree:
What's the relationship between `requirements.txt` and `setup.py`? Should `requirements.txt` even exist?

cournape:
Good question. The starting point is https://caremad.io/2013/07/setup-vs-requirement/

Now, there are exceptions to that rule, in particular, when you depend on dependencies not published on pypi (e.g. not OSS, or git commit), in which case requirements.txt are required. I think that used to apply, but does not anymore to satsolver (when we depended on enstaller master).

PR #175, opened 2016-04-01
ENH: Functions to compute dependencies and reverse dependencies

### LINKED ISSUE #171: Add functionality to find out dependencies and reverse dependencies (direct and transitive)
In the buildsystem, we currently rely on the following features derived from the old enstaller dependency solver:

1. Given a recipe (i.e. here a package name and its exact version), and a set of packages `S`, which subset of `S` depends on the recipe. E.g. you have the recipe for `numpy 1.10.3-2`, you want to find out which packages from `S` on that exact version of numpy. You want to find out either direct dependencies, or transitive ones. Usecase: when you update a package, say numpy, you want to have the set of packages that need testing to validate the new version is good.
2. Given a recipe and a set of packages `S`, which packages in `S` are dependencies of the recipe.

For both 1 and 2, we also want relaxed constraint on the recipe. The typical use case is when we update the version of a package (say numpy from `1.9.2` to `1.10.4`), we need to find every package that used to depend on `numpy 1.9.2`.

@johntyree, @noraderam  does that make sense ?

noraderam:
Closes #171 

`compute_dependencies` will return a set of packages on which packages for a given requirement depend.

`compute_reverse_dependencies` returns a set of packages which the requirement depends on.

Both can return either immediate dependencies or all recursive dependencies via the `transitive` flag.

@cournape, please let me know what you think

### DIFF
--- a/simplesat/compute_dependencies.py
+++ b/simplesat/compute_dependencies.py
@@ -20,7 +20,7 @@ def _compute_dependency_dict(pool, package_ids, transitive=False):
 
 def _reverse_mapping(mapping):
     reversed_map = defaultdict(set)
-    for key, vals in mapping.iteritems():
+    for key, vals in mapping.items():
         for v in vals:
             reversed_map[v].add(key)
 
--- a/simplesat/compute_dependencies.py
+++ b/simplesat/compute_dependencies.py
@@ -5,10 +5,11 @@ from simplesat.utils.graph import (package_lit_dependency_graph,
                                    transitive_neighbors)
 
 
-def _compute_dependency_dict(pool, package_ids):
+def _compute_dependency_dict(pool, package_ids, transitive=False):
     """ Return mapping of nodes to their dependents. """
-    graph = package_lit_dependency_graph(pool, package_ids, closed=False)
-    neighbors = transitive_neighbors(graph)
+    neighbors = package_lit_dependency_graph(pool, package_ids, closed=False)
+    if transitive:
+        neighbors = transitive_neighbors(neighbors)
     return neighbors
 
 
@@ -23,13 +24,17 @@ def _reverse_mapping(mapping):
 
 def _dependencies_for_requirement(pool, neighbor_mapping, requirement):
     result_dependencies = set()
-    for package in pool.what_provides(requirement):
-        package_id = pool.package_id(package)
+    for package_id in _package_ids_satisfying_requirement(pool, requirement):
         dependency_ids = neighbor_mapping[package_id]
         result_dependencies.update(pool.id_to_package(d_id)
                                    for d_id in dependency_ids)
 
     return result_dependencies
+
+
+def _package_ids_satisfying_requirement(pool, requirement):
+    for package in pool.what_provides(requirement):
+        yield pool.package_id(package)
 
 
 def _package_ids_from_repositories(pool, repositories):
@@ -44,6 +49,9 @@ def compute_dependencies(repositories, requirement, transitive=False):
     repositories : iterable of Repository objects
     requirement : Requirement
         The package requirement for which to compute dependencies.
+    transitive : bool
+        If True, recursively walk up the dependency graph. If False (default),
+        only returns the packages on which the package directly depends.
 
     Returns
     -----------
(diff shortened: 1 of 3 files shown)


noraderam:
Right, python 3...

The python 2 builds failed with 
```
Exception occurred:
  File "conf.py", line 118, in <module>
ImportError: No module named sphinx_rtd_theme
The full traceback has been saved in /tmp/sphinx-err-3KFe4d.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 1
The command "cd docs && make html" exited with 2.
```

Where did that come from?

cournape:
To test python 3, there is a tox file, so you can simply do, after installing detox:

``` bash
detox -e env1,env2
# or to run every version
detox
```

cournape:
As for the sphinx error,` sphinx_rtd_theme` is a separate package. It used to be a dependency of sphinx IIRC, maybe it isn't anymore ?

simplesat/compute_dependencies.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _package_ids_from_repositories(pool, repositories):
+    """ Return set of package ids in `pool` for all packages in the given repos
+    """
+    return set(pool.package_id(pkg) for repo in repositories for pkg in repo)
+
+
+def compute_dependencies(repositories, requirement, transitive=False):
+    """ Compute packages in the given repos on which `requirement` depends.
```
cournape: nitpick: can we have the public methods first  in the module ?

simplesat/compute_dependencies.py
```diff
@@ -0,0 +1,123 @@
...
+def _package_ids_satisfying_requirement(pool, requirement):
+    """ Yield package ids found in `pool` which satisfy `requirement`. """
+    for package in pool.what_provides(requirement):
+        yield pool.package_id(package)
+
+
+def _package_ids_from_repositories(pool, repositories):
+    """ Return set of package ids in `pool` for all packages in the given repos
+    """
+    return set(pool.package_id(pkg) for repo in repositories for pkg in repo)
```
cournape: ok, that function shows that we want to support iteration over a pool, and make it yield package ids :)

cournape:
One style comment, LGTM otherwise (we can implement pool iteration in another PR, as long as a ticket is created to track it).

noraderam:
> As for the sphinx error,sphinx_rtd_theme is a separate package. It used to be a dependency of sphinx IIRC, maybe it isn't anymore ?

Indeed--Sphinx 1.4 was released on March 28th; the changelog includes the comment "sphinx_rtd_theme has become optional. Please install it manually. ".

noraderam:
Travis has now been satisfied! @cournape, back to you

simplesat/compute_dependencies.py
```diff
@@ -20,7 +76,7 @@ def _compute_dependency_dict(pool, package_ids, transitive=False):
 
 def _reverse_mapping(mapping):
     reversed_map = defaultdict(set)
-    for key, vals in mapping.iteritems():
+    for key, vals in mapping.items():
```
cournape: you could use `six.iteritems` instead

cournape:
One trivial comment, LGTM otherwise

simplesat/compute_dependencies.py
noraderam: Done

cournape:
:+1:

simplesat/compute_dependencies.py
```diff
@@ -0,0 +1,125 @@
...
+    return graph
+
+
+def _reverse_mapping(mapping):
+    reversed_map = defaultdict(set)
+    for key, vals in six.iteritems(mapping):
+        for v in vals:
+            reversed_map[v].add(key)
+
+    return reversed_map
```
johntyree: I've been bitten a few times by forgetting to return a normal `dict` when using a `defaultdict` to build up a value.

:hamburger: :four_leaf_clover: :thought_balloon: ?


_I'm still working on my emoji-fluency._

PR #178, opened 2016-04-12
Add version

cournape:
This add `__version__` and a few other version-related metadata to the `simplesat` namespace.

This should work for both git repo and sdists.

@noraderam

### DIFF
--- a/setup_utils.py
+++ b/setup_utils.py
@@ -46,7 +46,6 @@ version_info = {version_info}
         except OSError:
             build_number = 0
 
-
     if _is_rc(version):
         release_level = "rc"
     elif not is_released:
--- a/setup_utils.py
+++ b/setup_utils.py
@@ -1,0 +1,149 @@
+import ast
+import os
+import re
+import subprocess
+
+
+_R_RC = re.compile("rc(\d+)$")
+
+
+def parse_version(path):
+    with open(path) as fp:
+        return _AssignmentParser().parse(fp.read())["version"]
+
+
+def write_version_py(filename, major, minor, micro, is_released,
+                     previous_version=None):
+    template = """\
+# THIS FILE IS GENERATED FROM SETUP_EXT
+version = '{final_version}'
+full_version = '{full_version}'
+git_revision = '{git_revision}'
+is_released = {is_released}
+
+msi_version = '{msi_version}'
+
+version_info = {version_info}
+"""
+    version = full_version = "{0}.{1}.{2}".format(major, minor, micro)
+    is_released = is_released
+
+    if not os.path.exists('.git'):
+        if os.path.exists(filename):
+            return
+        else:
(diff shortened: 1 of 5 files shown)


setup_utils.py
```diff
@@ -0,0 +1,149 @@
...
+_R_RC = re.compile("rc(\d+)$")
+
+
+def parse_version(path):
+    with open(path) as fp:
+        return _AssignmentParser().parse(fp.read())["version"]
+
+
+def write_version_py(filename, major, minor, micro, is_released,
+                     previous_version=None):
```
noraderam: Is the `previous_version` argument here just in case we need it later? It's not currently used.

setup_utils.py
```diff
@@ -0,0 +1,149 @@
...
+    git_rev = _git_version()
+    if previous_version is not None:
+        build_number = _compute_build_number("v" + previous_version)
+    else:
+        try:
+            out = _minimal_ext_cmd(['git', 'rev-list', '--count', 'HEAD'])
+            build_number = int(out.strip().decode('ascii'))
+        except OSError:
+            build_number = 0
+
```
noraderam: `flake8` complains about the extra blank line

noraderam:
I can confirm that this is essentially the same as the code in `jaguar`'s `setup.py`. How many other repos have equivalent functions? It would be nice if we had an identical approach across projects, complete with tests.

johntyree:
@noraderam I agree it would be good to standardize this. We do this pretty often, but always with just enough variation to break all of the little bespoke tools that get written each time to use it :(

setup_utils.py
cournape: Yes. The idea is to compute "build numbers" from some previous tag to avoid them increasing to 4-5 digit numbers. We don't have such a tag yet :)

cournape:
Agreed as well, this is why I put the meat  in a different file. I may take the time to make a small repo for it w/ basic instructions.

setup_utils.py
johntyree: When would our build numbers increase to 4-5 digit numbers?

cournape: Right now, the build number is calculated from the first commit. We're already at 704 for  enthought/satsolver, 994 for enthought/jaguar and 3700 for enthought/enstaller (and ~17000 for the buildsystem :) )

cournape: IIRC, I first added this because msi has a limit on the build number, though quite large (2**16 maybe)

johntyree: I'm still confused about how this works, but I assume you've thought about it much more than I have.

cournape: The idea is that from time to time, you "increment" the starting point for computing build numbers. By default, we start from the first commit, but later, we may start from tag `v0.2.0` for `v0.3.0`, etc...

PR #179, opened 2016-04-12
TST: update scenario name for marked test

johntyree:
This just updates two scenario files' names to reflect their purposes as marked package preservation tests.


noraderam:
LGTM

PR #186, opened 2016-04-27
BUG: fix #185.

### LINKED ISSUE #185: `add_package` is broken
Example:

``` python
package = PackageMetadata.from_egg("/home/davidc/.enpkg_eggs/rh5-64/okonomiyaki-0.14.1-1.egg")
repository = Repository()
repository.find_packages("okonomiyaki")
repository.add_package(package)
for p in repository:
    print(p)
```

It will iterate nothing because `add_package` assumes it is the only method that may create an entry in `Repository._name_to_repository`, but because `Repository._name_to_repository` is a default dict, (key, value) pairs may be created implicitly (as here in find_packages).

cournape:
@johntyree

### DIFF
--- a/simplesat/repository.py
+++ b/simplesat/repository.py
@@ -75,7 +75,7 @@ class Repository(object):
         If the same package is added multiple times to a repository, every copy
         will be available when calling find_package or when iterating.
         """
-        if package_metadata.name not in self._name_to_packages:
+        if len(self._name_to_packages[package_metadata.name]) == 0:
             bisect.insort(self._names, package_metadata.name)
 
         self._name_to_packages[package_metadata.name].append(package_metadata)
--- a/simplesat/repository.py
+++ b/simplesat/repository.py
@@ -61,7 +61,7 @@ class Repository(object):
 
     def __iter__(self):
         for name in self._names:
-            for package in self._name_to_packages.get(name, self._default_factory()):
+            for package in self._name_to_packages[name]:
                 yield package
 
     def add_package(self, package_metadata):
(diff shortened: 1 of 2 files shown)


simplesat/repository.py
```diff
@@ -75,7 +75,7 @@ def add_package(self, package_metadata):
         If the same package is added multiple times to a repository, every copy
         will be available when calling find_package or when iterating.
         """
-        if package_metadata.name not in self._name_to_packages:
+        if len(self._name_to_packages[package_metadata.name]) == 0:
```
johntyree: So the intention here *is* to insert new keys at this point and make it such that it makes no difference whether a key is empty or missing?  Is that correct?

cournape: Yes. The logic here is that `_names` is stored separately to keep iteration over `_name_to_packages` deterministic.

If you prefer, I can drop `defaultdict` and use a normal dict as per the discussion on the #edm channel

johntyree: I think it's alright. Just trying to make sure I understand how this fixes it.

Maybe a comment would help.

cournape: I converted to using a normal dict to see how painful it would be, and I think it makes the code clearer anyway :)

johntyree: :+1:

simplesat/repository.py
```diff
@@ -52,12 +52,13 @@ def __len__(self):
         )
 
     def __contains__(self, package_metadata):
-        return (
-            package_metadata in
-            self._name_to_packages.get(
-                package_metadata.name, self._default_factory()
+        if package_metadata.name in self._name_to_packages:
```
johntyree: Since we're not adding anything here, it really doesn't matter what the default factory is.

```python
candidates = self._name_to_packages.get(package_metadata.name, ())
return package_matadata in candidates
```
seems nicer to me.

Too late now. Don't change it.

PR #192, opened 2016-04-27
Release 0.2.1

cournape:


setup.py
```diff
@@ -7,7 +7,7 @@
 
 MAJOR = 0
 MINOR = 2
-MICRO = 1
+MICRO = 2
 
 IS_RELEASED = False
```
johntyree: `IS_RELEASED` ?

cournape: yes ?

cournape: oh, I think I see what you mean. With Simon, we figured it was very easy to forget to switch back `IS_RELEASED` to False, so we do the flip back and forth in one PR

johntyree: I'm confused. You mean this tagged commit is not what's released? You release it with this flag flipped and then flip it back?

cournape: yes, that's why there are 2 commits

johntyree: Somehow I didn't notice that. OK Sounds good :)

PR #193, opened 2016-04-27
BUG: fix package name parsing in requirement

johntyree:
As part of the lunch-n-learn, I made this change. I no longer remember why... **BUT** it *does* make the parsing of package names more consistent with the rest of the code both here and elsewhere.

The main change is that this permits a hyphen in between the name and the version and uses the regular expressions written to match the name and version as closely as possible.

For reasons still unknown, we actually had a test before to ensure that package names with a space instead of a hyphen caused errors. Since we don't allow spaces or hyphens in package names, I'm not sure why.

Thoughts on this?

### DIFF
--- a/simplesat/constraints/requirement.py
+++ b/simplesat/constraints/requirement.py
@@ -10,14 +10,18 @@ from simplesat.errors import (
 
 from .kinds import Any, Equal
 from .multi import MultiConstraints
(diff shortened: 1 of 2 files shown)


noraderam:
I don't have any problems with this approach.

Issue #177, opened 2016-04-12
"Obsoletes" and "Provides"


johntyree:
We want to think about how to deal with packages that provide other packages or virtual packages.

Problems include situations like `setuptools -> distribute -> setuptools` or `PIL -> pillow`.

Reference: https://www.debian.org/doc/debian-policy/ch-relationships.html#s-virtual

johntyree:
We have a first try at a spec for this written out. It needs to be added to the docs.

PR #194, opened 2016-04-27
Add "provides" package metadata.

johntyree:
This adds `PackageMetadata.provides` and the associated code around that. It works by dumping all of the providers of a given name into pool's the internal `name_to_packages` list. The rest of the behavior falls out from that.

If package A `provides` package B, then either A or B can be used to satisfy dependencies (or cause conflicts with) B. This does imply that package A *replaces* package B or that package B is somehow obsolete or that package A should be preferred. With the current implementation, if two packages provide the same thing then you'll get whichever has a higher version number, but this is unspecified. If you care about priority, you'll need to use some of the other metadata features (not yet implemented).

See https://github.com/enthought/sat-solver/issues/177

### DIFF
--- a/simplesat/constraints/tests/test_package_parser.py
+++ b/simplesat/constraints/tests/test_package_parser.py
@@ -226,17 +226,21 @@ class TestPrettyPackageStringParser(unittest.TestCase):
 
         # Given
         package_string = '; '.join((
-            "bokeh 0.2.0-3",
+            "bokeh_git 0.2.0-3",
             "install_requires (zope *, numpy ^= 1.8.0, requests >= 0.2)",
-            "conflicts (requests ^= 0.2.5, requests > 0.4, bokeh_git)",
+            "conflicts (requests ^= 0.2.5, requests > 0.4, bokeh)",
+            "provides (webplot ^= 0.1, bokeh)",
         ))
         r_install_requires = (
             ("numpy", (("^= 1.8.0",),)),
             ("requests", ((">= 0.2",),)),
             ("zope", (("*",),)))
         r_conflicts = (
-            ("bokeh_git", (('',),)),
+            ("bokeh", (('',),)),
             ("requests", (("^= 0.2.5", "> 0.4"),)))
+        r_provides = (
+            ("bokeh", (('',),)),
+            ("webplot", (("^= 0.1",),)))
 
         # When
         package = parse(package_string)
@@ -244,12 +248,14 @@ class TestPrettyPackageStringParser(unittest.TestCase):
         version = package['version']
         install_requires = package['install_requires']
         conflicts = package['conflicts']
+        provides = package['provides']
 
         # Then
-        self.assertEqual(name, "bokeh")
+        self.assertEqual(name, "bokeh_git")
         self.assertEqual(version, V("0.2.0-3"))
         self.assertEqual(install_requires, r_install_requires)
         self.assertEqual(conflicts, r_conflicts)
+        self.assertEqual(provides, r_provides)
 
 
 class TestPackagePrettyString(unittest.TestCase):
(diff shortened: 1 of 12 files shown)


simplesat/pool.py
```diff
@@ -47,7 +49,10 @@ def add_repository(self, repository):
             self._id += 1
             self._id_to_package_[current_id] = package
             self._package_to_id_[package] = current_id
-            self._packages_by_name_[package.name].append(package)
+            for constraints in package.provides:
+                req = Requirement.from_constraints(constraints)
+                assert not req.has_any_version_constraint
```
noraderam: Can you explain this assertion?

simplesat/package.py
```diff
@@ -85,15 +87,24 @@ def __init__(self, name, version, install_requires=None, conflicts=None):
...
                  ("six", (("> 1.2", "<= 1.2.3"), (">= 1.2.5-2",)))
-
         conflicts : tuple(tuple(str, tuple(tuple(str))))
             A tuple of tuples mapping distribution names to disjunctions of
             conjunctions of version constraints.
 
             This works the same way as install_requires, but instead denotes
             packages that must *not* be installed with this package.
+        provides : iterable of package names
+            The packages that are provided by this distribution. Useful when
```
noraderam: This makes it sounds like only the names of actual packages can be specified. From the examples, it looks like this could be an arbitrary label listed as a requirement. Is this purely a semantic difference?

simplesat/tests/test_rules_generator.py
```diff
@@ -59,6 +59,44 @@ def test_prefer_installed(self):
...
+        rules_generator = RulesGenerator(
+            pool, scenario.request,
+            installed_package_ids=installed_package_ids)
+        rules = list(rules_generator.iter_rules())
+
+        # Then
+        self.assertEqual(len(rules), 2)
+
+        # Given/When
+        conflict = rules[0]
```
noraderam: Is this actually a conflict?

noraderam:
Just a few questions

simplesat/pool.py
johntyree: Ah yes, I forgot to write about it. Really we need to move the entire package metadata spec into the docs here.

The long and short of it is that right now we don't have a good way to handle provides with version numbers. We might someday, but there aren't any good use-cases that I know of. This assertion is there to make sure package metadata authors haven't done that. It should probably be `ValueError` or `InvalidConstraint` or something instead. What do you think?

simplesat/package.py
johntyree: Indeed a package can `provide` anything it wants. If the provided name does not correspond to a real package somewhere, we call it a "virtual package." This nomenclature is borrowed from debian. I'll add that to the docstring somehow.

johntyree: In fact, this docstring is out of date, since the provides are also in the "constraint tuple" format.

simplesat/tests/test_rules_generator.py
johntyree: Nope. Copy :spaghetti:.

It's just a dependency relationship.

simplesat/pool.py
noraderam: This does sound like an `InvalidConstraint`. If this exception trickles up to users, we would want to make sure that the message was clear, too

simplesat/constraints/tests/test_package_parser.py
```diff
@@ -226,30 +226,36 @@ def test_complicated(self):
 
         # Given
         package_string = '; '.join((
-            "bokeh 0.2.0-3",
+            "bokeh_git 0.2.0-3",
             "install_requires (zope *, numpy ^= 1.8.0, requests >= 0.2)",
-            "conflicts (requests ^= 0.2.5, requests > 0.4, bokeh_git)",
+            "conflicts (requests ^= 0.2.5, requests > 0.4, bokeh)",
+            "provides (webplot ^= 0.1, bokeh)",
```
noraderam: How is this different from "provides with version numbers"?

johntyree: This is only testing our ability to parse, which separates the clauses and builds constraint tuples, but doesn't validate. In general we don't validate things up front due to the overhead of parsing version objects and the sheer number of packages. Also, we will not be parsing pretty strings during a normal work flow. IThe program which is using `simplesat` as a library will construct package metadata directly (which is why we guard against it below this level).

johntyree:
@noraderam :mailbox_with_mail:

noraderam:
LGTM

PR #195, opened 2016-04-28
ENH: api functions for simplifying and satisfying requests

johntyree:
These new API-level functions provide two abilities.

`simplify_requirements` - to turn a large list of package requirements into an equivalent set of smaller requirements for a given suite of packages. It does so by removing all requirements that are captured as dependencies of other requirements. For a self-consistent set of requirements, this will leave us with just the "leaf nodes" of the dependency tree. For requirements that are not self-consistent, the requirements with "mismatched" versions will also be included as they are not captured by any dependencies. This should make this a lossless operation.

`satisfy_requirements` - to use a set of packages to "solve" a requirement. This is the core functionality of this package, wrapped up in small function with an easy to use API. You pass in a set of packages and a set of requirements and it gives you back the subset of packages needed to satisfy them.

TODO:
- [x] test `satisfy_requirements`

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -11,7 +11,7 @@ from simplesat.request import JobType, Request
 from simplesat.rules_generator import RulesGenerator
 from simplesat.sat.policy import InstalledFirstPolicy
 from simplesat.sat import MiniSATSolver
-from simplesat.transaction import Transaction
+from simplesat.transaction import Transaction, InstallOperation
 from simplesat.utils import timed_context, connected_nodes
 
 
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -1,4 +1,5 @@
 import collections
+import itertools
 
 import six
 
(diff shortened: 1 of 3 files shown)


johntyree:
@noraderam this will be useful

simplesat/dependency_solver.py
```diff
@@ -142,6 +143,82 @@ def requirements_are_satisfiable(packages, requirements, modifiers=None):
...
+        The packages available to draw from when satisfying requirements.
+    requirements : list of Requirement
+        The requirements used to identify relevent packages.
+    modifiers : ConstraintModifiers, optional
+        If not None, modify requirements before resolving packages.
+
+    Returns
+    -------
+    tuple of PackageMetadata
+        Return a list of packages that together satisfy all of the
```
noraderam: Is it returning a list or a tuple?

simplesat/tests/test_solver.py
```diff
@@ -348,6 +351,87 @@ def test_missing_indirect_dependency_fails(self):
...
+        modifiers = ConstraintModifiers(
+            allow_newer=[u'A'],
+            allow_older=[u'B'],
+            allow_any=[u'C', u'D'],
+        )
+
+        # When
+        result = satisfy_requirements(
+            packages, requirements, modifiers=modifiers)
+        expected = tuple(packages[i] for i in (2, 7, 4, 6, 8))
```
noraderam: Why is the order important here?

simplesat/dependency_solver.py
```diff
@@ -142,6 +143,82 @@ def requirements_are_satisfiable(packages, requirements, modifiers=None):
...
+        `requirements`.
+
+    Raises
+    ------
+    SatisfiabilityError
+        If the `requirements` cannot be satisfied using the `packages`.
+
+    MissingInstallRequires
```
noraderam: Can you include a test for this failure mode?

simplesat/dependency_solver.py
```diff
@@ -142,6 +143,82 @@ def requirements_are_satisfiable(packages, requirements, modifiers=None):
...
+    needed_packages = packages_from_requirements(packages, requirements)
+    pool = Pool([Repository(packages)])
+    R = InstallRequirement.from_constraints
+    dependencies = set(itertools.chain.from_iterable(
+        pool.what_provides(R(con))
+        for package in needed_packages
+        for con in package.install_requires
+    ))
+    simple_requirements = requirements_from_packages(
+        package
```
noraderam: If order doesn't matter here, you could simplify this to 
```
    simple_requirements = requirements_from_packages(
        set(needed_packages).difference(dependencies)
    )
```

Not necessary, just a thought--it seems a little easier to parse, but that could be personal preference

noraderam:
That's all I've got. Back to you, @johntyree

simplesat/tests/test_solver.py
johntyree: The packages are in the order that the transaction has them, which, for installations, is  topological sorted.

johntyree: Comment about this added to docstring.

simplesat/dependency_solver.py
johntyree: Everything is using tuples. I'll fix the doc

simplesat/dependency_solver.py
johntyree: This failure mode is actually only possible when the solver is told to be "strict." That feature is meant for packages developers to have a sanity check before release.

I'll fix the doc and update the test.

simplesat/dependency_solver.py
johntyree: We don't care what the ordering is, but we do need it to be deterministic. In this case, we'd have to sort the output anyway to hide the unspecified iteration order of the set so we wouldn't really gain anything.

johntyree:
That's one thing we haven't talked about much yet. Be *really* careful when doing anything that iterates in an unspecified order. You can't let that behavior leak out to users. In practice this means a lot of sorting or using `OrderedDict.from(keys)` instead of `set`.

noraderam:
LGTM

PR #196, opened 2016-04-28
ENH: Use attributes for Request

noraderam:
Using attributes gives us equality comparison

### DIFF
--- a/simplesat/request.py
+++ b/simplesat/request.py
@@ -1,4 +1,4 @@
-from attr import attr, attributes
+from attr import attr, attributes, Factory
 from attr.validators import instance_of
 from enum import Enum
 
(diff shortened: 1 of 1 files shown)


cournape:
LGTM

simplesat/request.py
```diff
@@ -45,9 +46,8 @@ class Request(object):
     {'allow_older': [], 'allow_any': ['MKL'], 'allow_newer': []}
     """
 
-    def __init__(self, modifiers=None):
-        self.jobs = []
-        self.modifiers = modifiers or ConstraintModifiers()
+    modifiers = attr(default=Factory(ConstraintModifiers))
```
johntyree: It's possible to add validators to these as well. Not required of course, but see the `ConstraintModifiers` class if you're interested in how that works.

PR #197, opened 2016-04-29
REF: don't depend on okonomiyaki besides versions.

cournape:
* we remove any usage of okonomiyaki except for `okonomiyaki.versions` (whose API is stable).
* this makes it safe to remove the upper version bound for okonomiyaki dependency


cournape:
another trivial PR, merging to get `0.2.2` out.

PR #198, opened 2016-04-29
Backport #198 to 0.2.x

cournape:
REF: don't depend on okonomiyaki besides versions.
(cherry picked from commit 109e91287710900a2b57cbb83affc1130a3f7b87)

### DIFF
--- a/simplesat/constraints/tests/test_package_parser.py
+++ b/simplesat/constraints/tests/test_package_parser.py
@@ -1,7 +1,6 @@
 import sys
 import unittest
 
-from okonomiyaki.platforms import PythonImplementation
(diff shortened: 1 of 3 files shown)


simplesat/constraints/tests/test_package_parser.py
```diff
@@ -12,10 +11,6 @@
 from simplesat.errors import InvalidConstraint
 
 
-RUNNING_PYTHON = PythonImplementation(
```
johntyree: Why don't we need this anymore?

cournape: I don't think we ever needed it

PR #200, opened 2016-04-30
BUG: call to asdict must be deterministic

johntyree:
Without this, you can't do things like `yaml.dump(modifiers.asdict())` deterministically.


noraderam:
LGTM

PR #202, opened 2016-05-02
FIX: Don't set modifiers for non-InstallRqmnts

noraderam:
We do not want to apply modifiers to ConflictRequirement or Requirement
objects. We should not raise an exception in this case

@johntyree

### DIFF
--- a/simplesat/pool.py
+++ b/simplesat/pool.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import
 
 from .utils import DefaultOrderedDict
 from simplesat.constraints import (
-    ConstraintModifiers, Requirement, modify_requirement
+    InstallRequirement, Requirement, modify_requirement
 )
 from simplesat.errors import InvalidConstraint
 
@@ -85,7 +85,9 @@ class Pool(object):
 
     def modify_requirement(self, requirement):
         """Return requirement modified by the pool's ConstraintModifiers."""
-        if self.modifiers:
+        # It only makes sense for modifiers to apply to InstallRequirements,
+        # which are created from ``package.install_requires``
+        if self.modifiers and isinstance(requirement, InstallRequirement):
             requirement = modify_requirement(requirement, self.modifiers)
         return requirement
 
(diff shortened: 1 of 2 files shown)


johntyree:
The reason it was written this way is because it's very easy to use an incorrect requirement type when implementing something new. Having it raise an exception catches this immediately, preventing you from accidentally modifying (or not modifying) a requirement in a way that doesn't make sense.

I think the code paths for the various kinds of requirements should be entirely different such that `modify_requirement` isn't called on the wrong type in the first place. Having it silently pass through and while doing nothing seems like the wrong approach to me because if there's an error, it won't be obvious until much later, if at all.

How did you bump into this?

johntyree:
We are taking different approaches to the "Don't modify the wrong type of requirements" problem. The way it's currently written in master (with the exception) puts the onus on the caller of `what_provides` to know whether or not the requirement is modifiable. Your PR makes it so the onus is on the creator of the Requirement object. On the one hand, I think it's most philosophically gratifying to say that the code that creates the Requirement object should choose the right one and no one else should worry about it. On the other hand, it's pretty easy to get that wrong and no one will tell you about it.

The exception does what it is supposed to as far as catching misuses of `what_provides`, BUT the if requirement is the correct one then this PR's way will also do the intended thing. The misuse is that the pool defaults to modifiying, but that these requirements are job requirements from the user and should not be modified. Personally I prefer the exception way because it's harder to do it wrong and easier to figure out when you do. I think this is useful when implementing something new.

noraderam:
@johntyree OK, I think I understand what you mean now. There's actually a `use_modifiers` keyword argument for `what_provides`. We ought to be using that for `Requirement`s.

noraderam:
@johntyree Thanks for the explanation. I zoomed in a bit too quickly to a solution and ignored the larger context. I'd like to close this PR and instead make the fix elsewhere, in the code which calls `what_provides`.

johntyree:
@noraderam right. So the two approaches are either to explicitly tell the pool that you don't want to use modifiers, or to let the fact that you didn't pass it an `InstallRequirement` implicitly convey that information.

Either is valid and will work, but I think raising the exception is safer, at least this early in the game while we're still implementing core features.

PR #209, opened 2016-05-05
Bug/asserted requirement

johntyree:
This fixes a bug that crashes the `UndeterminedClausePolicy` when the solver is run with a package name that has only one corresponding package version. This gets turned into a rule with one literal, which is then turned into a clause with one literal.

Clauses with one literal are assertions that warrant immediate assignment of a value. Since we knew that they were satisfied, we weren't tracking these clauses at all. This leads to problems in the Policy, which assumes that all package literals will appear in at least one clause.

We didn't run into this sooner because it requires *both* a requirement for a package with a single version *and* a requirement for some other package having more than one version such that the solver is forced to ask the policy for a suggestion.

The fix is to also track clauses with one literal. Clauses with no literals are still ignored.

@noraderam this one is subtle but worth understanding.

### DIFF
--- a/simplesat/sat/minisat.py
+++ b/simplesat/sat/minisat.py
@@ -339,7 +339,9 @@ class MiniSATSolver(object):
             # Clause is guaranteed to be false under the current variable
             # assignments.
             self.status = False
-        elif len(clause) == 1:
+            return
+
+        if len(clause) == 1:
             # Unit facts are enqueued.
             if not self.enqueue(clause[0], cause=clause):
                 # Bail out if we've found a conflict
@@ -353,7 +355,7 @@ class MiniSATSolver(object):
             self.watches[-p].append(clause)
             self.watches[-q].append(clause)
 
-            self.clauses.append(clause)
+        self.clauses.append(clause)
 
     def _setup_assignments(self):
         """Initialize assignments table.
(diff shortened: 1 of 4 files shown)


noraderam:
> The fix is to also track clauses with one literal.

Does this have any performance implications?

johntyree:
@noraderam absolutely! Some problems that were previously unsolvable can now run to completion :)

Jokes aside, some things are linear in the number of clauses but I don't think it gets any worse than that. This shouldn't have any effect on the search itself and it's almost certainly cheaper than wrapping exception handlers around the dictionary lookups where this bug lives.

noraderam:
Let me see if I understand this.
For the clauses with only one literal, we automatically made the assignment, but we didn't track the clause. When the policy is asked to suggest a new literal, it reviews the existing assignments. The error arises because it's trying to find clauses which contain the literal we assigned and none are tracked.

The change means that the policy will have these clauses with only one literal. Thus, no `KeyError` is raised and we can proceed as expected. We aren't changing any of the actual content of the satisfiability problem, so there's no reason to expect any negative side effects.

johntyree:
Si.

The reason that this hadn't come up before is two fold.

1. If a problem can be solved entirely by unit propagation then the policy is not used at all. For example, if you ask to install `jsonschema` and there's only one version, then we know immediately what to do.
2. If the package has any dependencies or reverse-dependencies, then even when we don't track the single-literal unit clause, there will still be others capturing the dependency rules and so `_id_to_clauses` won't be empty.

noraderam:
LGTM

PR #219, opened 2016-05-26
Function to compute some minimal unsatisfiable subsets

cournape:
@noraderam

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -3,6 +3,7 @@ import itertools
 
 import six
 
+from simplesat.constraints import ConstraintModifiers
 from simplesat.constraints.requirement import InstallRequirement
 from simplesat.errors import NoPackageFound, SatisfiabilityError
 from simplesat.pool import Pool
@@ -130,7 +131,9 @@ def requirements_are_satisfiable(packages, requirements, modifiers=None):
     bool
         Return True if the `requirements` can be satisfied by the `packages`.
     """
-    request = Request()
+    modifiers = modifiers or ConstraintModifiers()
+
+    request = Request(modifiers=modifiers)
     for requirement in requirements:
         request.install(requirement)
     repositories = (Repository(packages),)
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -262,7 +262,7 @@ def minimal_unsatisfiable_subset(clauses, callback):
     this property.
     """
     clauses = tuple(clauses)
-    if callable(clauses):
+    if callback(clauses):
         raise UnexpectedlySatisfiable()
 
     def split(s):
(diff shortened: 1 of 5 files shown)


simplesat/dependency_solver.py
```diff
@@ -216,6 +221,87 @@ def simplify_requirements(packages, requirements):
...
+
+    We do a binary search on the clauses by splitting them in halves A and B.
+    If A or B is UNSAT, we use that and repeat. Otherwise, we recursively check
+    A, but each time we do a sat query, we include B, until we have a minimal
+    subset A* of A such that A* U B is UNSAT. Then we find a minimal subset B*
+    of B such that A* U B* is UNSAT. Then A* U B* will be a minimal
+    unsatisfiable subset of the original set of clauses.
+
+    Proof: If some proper subset C of A* U B* is UNSAT, then there is some
+    clause c in A* U B* not in C. If c is in A*, then that means (A* - {c}) U
```
noraderam: I found the use of `c` here confusing--I initially read this as `c \in C`. Could you use a different letter?

simplesat/dependency_solver.py
```diff
@@ -216,6 +221,87 @@ def simplify_requirements(packages, requirements):
...
+    B* is UNSAT, and hence (A* - {c}) U B is UNSAT, since it is a superset,
+    which contradicts A* being the minimal subset of A with such property.
+    Similarly, if c is in B, then A* U (B* - {c}) is UNSAT, but B* - {c} is a
+    strict subset of B*, contradicting B* being the minimal subset of B with
+    this property.
+    """
+    clauses = tuple(clauses)
+    if callable(clauses):
```
noraderam: This should be `callback`, not `callable`. It works because `callable` is a built-in, but it's not actually making the desired check.

There should be a test that makes sure we raise `UnexpectedlySatisfiable` when expected

noraderam:
Can you give an example of a minimal set with more than two elements?

noraderam:
> Can you give an example of a minimal set with more than two elements?

There are certainly simpler answers, but here's what I came up with:

```
        A 1.0-1; provides (X)
        B 1.0-1; provides (X, Y, Z)
        C 1.0-1; provides (Y)
        D 1.0-1; provides (Y, Z)
        P 1.0-1; depends (X)
        Q 1.0-1; depends (Y); conflicts (A ^= 1.0)
        R 1.0-1; depends (Z); conflicts (B ^= 1.0)
```
```
        requirements = [
            Requirement._from_string("P"),
            Requirement._from_string("Q"),
            Requirement._from_string("R"),
```

`P, Q, R` is a minimal unsatisfiable set

I think it would be worthwhile to include a test for a case like this

cournape:
@noraderam how does that look now ?

noraderam:
LGTM

PR #220, opened 2016-05-26
ENH: Optionally ignore installed packages 

noraderam:
Motivation:
The initial approach for `soft_update` was to remove those packages from the `installed_repository` passed to the `policy`. Reducing the number of knobs on `DependencySolver` to turn can increase our confidence in use of the solver. Instead of editing the `installed_repository` in a somewhat hacky manner we can include all necessary information in the `Request`.

Changes:
- A new job type `soft_update` is added to `Request`. The policy will prefer the installed versions of all packages except those which have `soft_update` jobs. The existing `update` job is renamed to `hard_update`

- Remove policy as attribute for `DependencySolver`. The `policy` needs to be created with the same `pool` and `installed_repository` as the `DependencySolver`. It strikes me as safer to create the policy when we need it rather than allowing the user to pass it in. This also allows us to customize the policy based on the request. If different policies are added at a later date we can consider the best way to add back this functionality.

- Remove `prefer_installed` flag from `policy`. The `policy` is now created by the `DependencySolver` and is not customizable by the user. The same behavior can be reached by changing the `request`.

@johntyree I'd love to hear your opinion

@cournape Please take a look

### DIFF
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -322,9 +322,9 @@ class DependencySolver(object):
         pool = self._pool
         installed_repository = self._installed_repository
 
-        policy = InstalledFirstPolicy(pool, installed_repository)
         all_requirement_ids = []
 
+        soft_update_packages = set()
         for job in request.jobs:
             assert job.kind in (
                 JobType.install, JobType.remove, JobType.hard_update,
@@ -343,15 +343,22 @@ class DependencySolver(object):
                 def key(package):
                     return (package.version, package in installed_repository)
                 providers = [max(providers, key=key)]
+            elif job.kind == JobType.soft_update:
+                soft_update_packages.update((pkg for pkg in providers
+                                            if pkg in installed_repository))
 
-            requirement_ids = [pool.package_id(p) for p in providers]
-            policy.add_requirements(requirement_ids)
-            all_requirement_ids.extend(requirement_ids)
+            all_requirement_ids.extend([pool.package_id(p) for p in providers])
 
         installed_package_ids = collections.OrderedDict()
         for package in installed_repository:
             package_id = pool.package_id(package)
             installed_package_ids[package_id] = package
+
+        # Prefer the installed versions of all packages
+        policy = InstalledFirstPolicy(
+            pool, installed_repository,
+            ignore_installed_packages=soft_update_packages)
+        policy.add_requirements(all_requirement_ids)
 
         rules_generator = RulesGenerator(
             pool, request, installed_package_ids=installed_package_ids,
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -233,9 +233,6 @@ class DependencySolver(object):
         Repositories containing package available for installation.
     installed_repository : Repository
         Repository containing the packages which are currently installed.
-    policy : Policy, optional
-        The policy for suggesting new packages during the search phase. If none
-        is given, then ``simplsat.policy.InstalledFirstPolicy`` is used.
     use_pruning : bool, optional
         When True, attempt to prune package operations that are not strictly
         necessary for meeting the requirements. Without this, packages whose
@@ -266,7 +263,7 @@ class DependencySolver(object):
     """
 
     def __init__(self, pool, remote_repositories, installed_repository,
-                 policy=None, use_pruning=True, strict=False):
+                 use_pruning=True, strict=False):
         self._pool = pool
         self._installed_repository = installed_repository
         self._remote_repositories = remote_repositories
@@ -276,10 +273,6 @@ class DependencySolver(object):
 
         self.strict = strict
         self.use_pruning = use_pruning
-
-        self._policy = policy or InstalledFirstPolicy(
-            pool, installed_repository
-        )
 
     def solve(self, request):
         """Given a request return a Transaction that would satisfy it.
@@ -302,11 +295,12 @@ class DependencySolver(object):
         modifiers = request.modifiers
         self._pool.modifiers = modifiers if modifiers.targets else None
         with self._last_rules_time:
-            requirement_ids, rules = self._create_rules_and_initialize_policy(
+            init_rules_and_policy = self._create_rules_and_initialize_policy
+            requirement_ids, rules, policy = init_rules_and_policy(
                 request
             )
         with self._last_solver_init_time:
-            sat_solver = MiniSATSolver.from_rules(rules, self._policy)
+            sat_solver = MiniSATSolver.from_rules(rules, policy)
         with self._last_solve_time:
             solution = sat_solver.search()
         solution_ids = _solution_to_ids(solution)
@@ -328,6 +322,7 @@ class DependencySolver(object):
         pool = self._pool
         installed_repository = self._installed_repository
 
+        policy = InstalledFirstPolicy(pool, installed_repository)
         all_requirement_ids = []
 
         for job in request.jobs:
@@ -349,7 +344,7 @@ class DependencySolver(object):
                 providers = [max(providers, key=key)]
 
             requirement_ids = [pool.package_id(p) for p in providers]
-            self._policy.add_requirements(requirement_ids)
+            policy.add_requirements(requirement_ids)
             all_requirement_ids.extend(requirement_ids)
 
         installed_package_ids = collections.OrderedDict()
@@ -361,7 +356,7 @@ class DependencySolver(object):
             pool, request, installed_package_ids=installed_package_ids,
             strict=self.strict)
 
-        return all_requirement_ids, list(rules_generator.iter_rules())
+        return all_requirement_ids, list(rules_generator.iter_rules()), policy
 
 
 def _connected_packages(solution, root_ids, pool):
--- a/simplesat/dependency_solver.py
+++ b/simplesat/dependency_solver.py
@@ -347,7 +347,7 @@ class DependencySolver(object):
                 soft_update_packages.update((pkg for pkg in providers
                                             if pkg in installed_repository))
 
-            all_requirement_ids.extend([pool.package_id(p) for p in providers])
+            all_requirement_ids.extend(pool.package_id(p) for p in providers)
 
         installed_package_ids = collections.OrderedDict()
         for package in installed_repository:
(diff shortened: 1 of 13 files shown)


simplesat/dependency_solver.py
```diff
@@ -342,26 +338,33 @@ def _create_rules_and_initialize_policy(self, request):
...
         installed_package_ids = collections.OrderedDict()
         for package in installed_repository:
             package_id = pool.package_id(package)
             installed_package_ids[package_id] = package
 
+        # Prefer the installed versions of all packages
+        policy = InstalledFirstPolicy(
+            pool, installed_repository,
+            ignore_installed_packages=soft_update_packages)
+        policy.add_requirements(all_requirement_ids)
```
cournape: we are now `passing all_requirements_ids` instead of `[pool.package_id(p) for p in providers]` to `policy.add_requirements`. What are the consequences (if any) ?

noraderam: Previously we called `self._policy.add_requirements(requirement_ids)` at each iteration. We now build up a list of the same requirement ids (`all_requirement_ids.extend([pool.package_id(p) for p in providers])`) and add them all at once. There is no change in the actual ids added; as far as I can tell, this behaves exactly the same.

simplesat/request.py
```diff
@@ -8,7 +8,8 @@
 class JobType(Enum):
     install = 1
     remove = 2
-    update = 3
+    soft_update = 3
```
cournape: This makes 0.4.0 backward incompatible. This is fine, but can you make 2 small PRs on  edm master and maintenance/0.6.x to lock the current satsolver versions to `>= 0.3.1, < 0.4.0` ?

simplesat/rules_generator.py
```diff
@@ -495,11 +495,11 @@ def _add_installed_package_rules(self, package):
 
     def _add_job_rules(self):
         for job in self.request.jobs:
-            if job.kind == JobType.install:
+            if job.kind == JobType.install or job.kind == JobType.soft_update:
```
cournape: w/ enum, we can just do `if job.kind in (JobType.install, JobType.soft_update)`

simplesat/sat/policy/undetermined_clause_policy.py
```diff
@@ -16,17 +15,23 @@ class UndeterminedClausePolicy(IPolicy):
...
         self._pool = pool
-        self.prefer_installed = prefer_installed
+
         by_version = six.functools.partial(pkg_id_to_version, self._pool)
-        self._installed_ids = sorted(
-            (pool.package_id(package) for package in installed_repository),
-            key=by_version
-        )
-        self._preferred_package_ids = {
+        installed_packages = set(package for package in installed_repository)
```
cournape: `installed_packages = set(installed_repository)` should be equivalent

noraderam: Oops! Refactored this too many times to notice that I was now not actually doing anything here

simplesat/dependency_solver.py
cournape: oh, right, I missed the loop while looking at the GH diff.

simplesat/dependency_solver.py
```diff
@@ -342,26 +338,33 @@ def _create_rules_and_initialize_policy(self, request):
...
                     return (package.version, package in installed_repository)
                 providers = [max(providers, key=key)]
+            elif job.kind == JobType.soft_update:
+                soft_update_packages.update((pkg for pkg in providers
+                                            if pkg in installed_repository))
 
-            requirement_ids = [pool.package_id(p) for p in providers]
-            self._policy.add_requirements(requirement_ids)
-            all_requirement_ids.extend(requirement_ids)
+            all_requirement_ids.extend([pool.package_id(p) for p in providers])
```
cournape: building an actual list is not needed I think

simplesat/sat/policy/undetermined_clause_policy.py
```diff
@@ -16,17 +15,23 @@ class UndeterminedClausePolicy(IPolicy):
...
+        prefer_installed = installed_packages - ignore_installed_packages
```
cournape: we can rename `prefer_installed` to `preferred_installed_packages` ? Seems more consistent

simplesat/sat/policy/undetermined_clause_policy.py
```diff
@@ -89,13 +93,12 @@ def get_next_package_id(self, assignments, clauses):
...
-            # NOTE: when repository priority is implemented, this will cause
-            # the virtual <installed> repository to act as if it always has the
-            # highest priority.
-            key = self._package_key(candidate_id)
-            candidate_id = self._preferred_package_ids.get(key, candidate_id)
+        # If this exact package version is available locally, use it.
+        # NOTE: when repository priority is implemented, this will cause
+        # the virtual <installed> repository to act as if it always has the
+        # highest priority.
+        key = self._package_key(candidate_id)
```
cournape: since we now always execute those lines, what is the difference compared to before (where we did not execute them by default) ?

cournape:
nitpicks, and one logic question

simplesat/sat/policy/undetermined_clause_policy.py
noraderam: Yeah... I was sacrificing name for line length. I forgot that in these repos we use a different line length standard, so it doesn't matter anyway

simplesat/sat/policy/undetermined_clause_policy.py
noraderam: Previously, if `prefer_installed` were `True`, we would default to suggesting the `package_id` of installed packages, which come from the `installed_repository`. This block made sure that if we were _not_ using `prefer_installed` we would still get the package from the installed repo rather than the remote one--it prevented an unnecessary `remove`, `install` step to get the same version of a package.

With the new changes, we always call this block to make sure that we use the installed repo if possible. There is no change in behavior. We could put this into a `if candidate_id not in `self.prefer_installed_pkg_ids` block, but that seemed like an extra bit of logic that doesn't gain us much in terms of speed

cournape:
:+1:

simplesat/sat/policy/undetermined_clause_policy.py
noraderam: 😞  I'm hitting an infinite loop with this when testing a few more complicated cases. I think it has to do with how we suggest candidates. Investigating...

noraderam: Ok, here's the problem. My understanding was incomplete--this does what I thought most of the time, but it can also cause a huge problem.

The policy came up with a candidate to propose. We hijack that here and suggest a different (equivalent) candidate. The next time around, the policy may try to suggest the original candidate again and we get stuck in an infinite loop. This would have been possible on `master`, too, but would not come up through our normal way of calling the solver because `prefer_installed` was set to `True`.

noraderam: If the block is removed, you can end up removing the installed package and then re-installing same version from the remote.
It may be best to remove this block and add some post-processing to ensure that we use the installed versions if possible.

noraderam:
@cournape I've added a fix with an accompanying regression test

cournape:
:+1:

cournape:
@noraderam once this and #218 are in master, can you take care of releasing `0.4.0` or do you prefer for me to take care of it ?

noraderam:
> once this and #218 are in master, can you take care of releasing 0.4.0 or do you prefer for me to take care of it ?

I'll plan to take care of the release and will let you know if I run into any issues

simplesat/transaction.py
```diff
@@ -113,23 +121,40 @@ def _safe_operations(self, pool, decisions, installed_package_ids):
...
-            operations.append(RemoveOperation(package))
+        install_operations = []
+        remove_operations = []
 
         # Installations should happen bottom up
         for package_id in installs:
-            package = pool.id_to_package(package_id)
-            operations.append(InstallOperation(package))
+            # The solver may suggest removing a currently installed package and
+            # installing another version of that same package from a different
```
johntyree: I haven't gotten it straight in my head yet how we handle identical looking packages from different repositories. We've already said that they aren't equivalent, but we often assume that any attempt to swap one for the other must be a mistake. Do we have the behavior written down anywhere? @noraderam @cournape

noraderam: At the moment, we don't exactly "handle" this. We do treat them as equivalent: the policy and solver don't care about the repository info, just the package name and version. In a discussion on [Slack](https://enthought.slack.com/archives/edm/p1464364546004725) @cournape noted, "Even if we implement repository priority at a latter time, we will still always prefer installed over non installed for a same `(name, version)` pair."

https://github.com/enthought/sat-solver/pull/220/commits/b8aedbb9376103b39886da7a04fc89f79b76f7ba describes the reason for this change. It's not the only way to do this but it was easier than overhauling more of the policy to escape the infinite loop.

johntyree: I understand the reason for the change and this isn't the first time it has been fixed. I just remember a discussion where we said that packages were identified by (name, version, repo), not just (name, version). In practice, though, we seem to explicitly ignore it.

noraderam: As I currently understand it, we pass around the repo information but we don't actually do anything with it. The only place where it matters is when considering which packages are currently installed (available in the installed repo). This would be relevant if we implemented repo prioritization. Can you think of any other situations when we would care? Specifically, what would it mean if those identical-looking packages from different repositories are not actually equivalent?

johntyree: It could mean that two teams that don't know anything about each other have published packages with the same name. The repos used within a single run don't have to come from a single universe, right? We might have two groups in a bank that have both pushed a `weekly_reports` package to their respective repositories, for example. Now a third-party user can't pull packages from both repos without running these identically named packages through the solver. So what happens then?

noraderam: Oooh, interesting. That opens up a whole scope of questions. How does the user specify which `weekly_reports` package they want? How do other packages refer to it when specifying dependencies and conflicts? Can you get both `weekly_reports`? Do we really want to have to handle all of that?

johntyree: I thought we were handling it via repo priority, but I don't see how without going back and changing all the places where we assume, sometimes implicitly, that the repos don't matter. That's why I'm asking if we've made a decision about it.

@cournape are any of the things nora's asking about on the road map? Do we have a plan for packages that only differ by repository info?

PR #221, opened 2016-05-27
BUG: fix index.json -> yaml script.

cournape:
@noraderam

### DIFF
--- a/scripts/index_to_yaml.py
+++ b/scripts/index_to_yaml.py
@@ -1,8 +1,10 @@
 import argparse
 import collections
+import json
 import operator
 import sys
 
+import six
 import yaml
 
 from okonomiyaki.versions import EnpkgVersion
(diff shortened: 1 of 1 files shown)


scripts/index_to_yaml.py
```diff
@@ -1,8 +1,10 @@
 import argparse
 import collections
+import json
 import operator
```
noraderam: `operator` is no longer used

scripts/index_to_yaml.py
```diff
@@ -12,10 +14,63 @@
...
+        json_dict = json.load(fp)
+
+    repository = Repository()
+    repository_info = RepositoryInfo("remote")
+
+    for key, entry in six.iteritems(json_dict):
+        raw_name = key.split("-")[0].lower()
+        version_string = "{0}-{1}".format(entry["version"], entry["build"])
+        version = EnpkgVersion.from_string(version_string)
+        name_to_install_requires =  parse_index_packages_to_install_requires(
```
noraderam: Nit: Two spaces after operator

scripts/index_to_yaml.py
```diff
@@ -12,10 +14,63 @@
...
+            if package_version.build == 0:
+                constraint = "^= {0}".format(version_part)
+            else:
+                constraint = "== {0}".format(version_part)
+
+        name_to_install_requires[name].append(constraint)
+
+    return dict(
+        (name, tuple(value))
+        for name, value in six.iteritems(name_to_install_requires)
```
noraderam: You iterate over the entries to create a `dict` and then immediately iterate over the `dict` to create a `tuple`. Why not just create the form that you want here?

PR #227, opened 2016-06-01
Maint/start 0.5.x

noraderam:


cournape:
Oups, I did not see that one, and made another one :(

PR #229, opened 2016-06-18
ENH: support ConstraintsModifiers.remove

### LINKED ISSUE #214: Add `remove` method to `constraints.ConstraintModifiers`
Method can use the `update_difference` set method under the covers.

johntyree:
Raises a `TypeError` if a string is passed instead of an iterable of strings.

Closes #214

### DIFF
--- a/simplesat/constraints/constraint_modifiers.py
+++ b/simplesat/constraints/constraint_modifiers.py
@@ -96,6 +96,21 @@ class ConstraintModifiers(object):
         self.allow_newer.update(other_modifiers.allow_newer)
         self.allow_older.update(other_modifiers.allow_older)
 
+    def remove(self, packages):
+        """ Remove all modifiers related to `packages`.
+
+        Parameters
+        ----------
+        packages : a list, tuple, or set of strings
+            The package names that should be completely removed.
+        """
+        disallowed = (type(b""), type(u""))
+        if isinstance(packages, disallowed):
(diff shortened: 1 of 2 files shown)


simplesat/constraints/constraint_modifiers.py
```diff
@@ -96,6 +96,21 @@ def update(self, other_modifiers):
...
+    def remove(self, packages):
+        """ Remove all modifiers related to `packages`.
+
+        Parameters
+        ----------
+        packages : an iterable of strings
+            The package names that should be completely removed.
+        """
+        disallowed = (type(b""), type(u""))
+        if isinstance(packages, disallowed):
```
cournape: shouldn't we instead check for iterable of strings ?

Also, iterable is a bit of a misnomer here, as it will break w/ a non sequence. If we want to make it work for iterables, we should first "reify" the iterable before passing it down.

johntyree: Yes. I missed the reuse. Good catch.

noraderam:
LGTM

simplesat/constraints/constraint_modifiers.py
johntyree: We check for strings because it's easy to accidentally pass them instead of collections, which is bad.

`constraint.remove(('foo'))` would not do what you want.

johntyree:
Don't mind me, just working on an open source project in my free time. :suspect:

PR #230, opened 2016-06-21
ENH: ConstraintModifiers can now accept nested iterables in its constructor

prashantmital:


noraderam:
@prashantmital What's the motivation for this change?

noraderam:
@prashantmital I'd be more in favor of handling this when building the `ConstraintModifiers` object. This doesn't feel like something that `ConstraintModifiers` should actually have to handle itself.

prashantmital:
@noraderam that makes sense. I'll close this.

PR #231, opened 2016-06-22
ENH: return error message when running checks for satisfiability 

prashantmital:


prashantmital:
While this approach is simple enough, the need to index into the returned value to get a `True`/`False` result is less than ideal IMO. Instead we could have a class equipped with the `__bool__` (`__nonzero__` for Python < 2) method that stores the error message as one of its members. What do you think?

noraderam:
How about we split the difference and use a named tuple?

noraderam:
I'm confused about the test grouping: What is tested in `test_solver` and what is tested in `test_dependency_solver`?

prashantmital:
I'll need to take a closer look to figure that out. I just *fixed* the pre-existing tests without paying too much attention to the organization.

noraderam:
It looks like `test_dependency_solver` was added in #219, likely because it wasn't clear that tests already existed in `test_solver`

noraderam:
I'm happy with how this looks, assuming that tests pass

prashantmital:
@noraderam Safe to merge?

cournape:
Can we rename `Result.result` to `Result.is_satisfiable` or something similar ?

noraderam:
LGTM

PR #244, opened 2016-07-20
MAINT: Release 0.6.0

noraderam:


prashantmital:
LGTM.

Side question: is there anyway to see tags in a PR? I imagine the first commit here has the `v0.6.0` tag on it, but I can't figure out a way of confirming that.

noraderam:
> Side question: is there anyway to see tags in a PR? I imagine the first commit here has the v0.6.0 tag on it, but I can't figure out a way of confirming that.

I don't think there's a way to see that. However, in this case, the tag does not exist yet as I was waiting for sign off before creating the tag.

prashantmital:
👍

PR #247, opened 2016-08-05
ENH: Add method to compute leaf packages.

ashishbhate:


simplesat/compute_dependencies.py
```diff
@@ -63,6 +63,42 @@ def compute_reverse_dependencies(repositories, requirement, transitive=False):
...
+    Parameters
+    ----------------
+    repositories : iterable of Repository objects
+
+    Returns
+    -----------
+    dependencies : set
+        Set of leaf packages in the given repositories.
+    """
+    pool = Pool(repositories)
```
noraderam: This first bit is shared with `compute_reverse_dependencies`--they should use a common utility

simplesat/compute_dependencies.py
```diff
@@ -63,6 +63,42 @@ def compute_reverse_dependencies(repositories, requirement, transitive=False):
...
+    package_ids = set(pool.iter_package_ids())
+    neighbors = _compute_dependency_dict(pool, package_ids, transitive=False)
+
+    # Reverse mapping so that package ids point to the packages which depend
+    # on them
+    reverse_neighbors = _reverse_mapping(neighbors)
+
+    leaf_packages = set()
+    for repository in repositories:
```
noraderam: You could do `for package in pool.iter_packages`

simplesat/compute_dependencies.py
```diff
@@ -63,6 +63,42 @@ def compute_reverse_dependencies(repositories, requirement, transitive=False):
...
+        for package in repository:
+            package_string = package.name + "-" + str(package.version)
```
noraderam: It seems like you ought to be able to create the `InstallRequirement` directly from the `PackageMetadata`, but I can't find a way to do so.

simplesat/tests/test_compute_dependencies.py
```diff
@@ -109,3 +110,21 @@ def test_chained_dependencies_transitive(self):
...
+
+
+class TestComputeLeafPackages(unittest.TestCase):
+
+    def setUp(self):
+        repo_0 = Repository(packages_from_definition(PACKAGE_DEF_0))
+        repo_1 = Repository(packages_from_definition(PACKAGE_DEF_1))
+        self.repos = [repo_0, repo_1]
+
+    def test_simple(self):
```
noraderam: Can you make sure to cover the case where the same package is available in multiple repositories?

ashishbhate:
Updated

PR #260, opened 2017-03-13
Fix 7z requirement parsing error

JCorson:
I am not terribly crazy about this fix, but it is the best I have come up with so far. The main issue seems to be that if a package name starts with a number, it is parsed as a `VersionToken`. The difficulty is that we actually want some valid version constraint strings to also be valid package names. This fix just adds a check to ensure the distribution does not actually match the `_DISTRIBUTION_NAME_R` pattern before raising the error.

### DIFF
--- a/simplesat/constraints/parser.py
+++ b/simplesat/constraints/parser.py
@@ -201,7 +201,9 @@ class _RawRequirementParser(object):
             if len(requirement_block) == 3:
                 distribution, operator, version = requirement_block
                 if not isinstance(distribution, DistributionNameToken):
-                    raise InvalidConstraint(msg + ' (bad distirbution name)')
+                    if re.match(_DISTRIBUTION_R, distribution.value) is None:
+                        raise InvalidConstraint(
+                            msg + ' (bad distirbution name)')
                 if not isinstance(version, VersionToken):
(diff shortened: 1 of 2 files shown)


cournape:
@JCorson why not fixing the `_R_DISTRIBUTION_NAME` regex instead ?

JCorson:
@cournape I *think* I may have come up with a better way. The issue is more that the `_VERSION_R` is matching package names that start with a number than an issue with `_DISTRIBUTION_NAME_R` itself. https://github.com/enthought/sat-solver/pull/260/commits/1d25be67bc45979686cf5cb1aa475c6ea0f8c9ac changes `_VERSION_R` to only match if the group ends with a word boundary and is not followed by a space. Does that make sense for this case?

johntyree:
I think pinning the `_VERSION_R` to the end of the string is pretty sensible. :+1:

cournape:
@johntyree long time no see ;) Since you're here, I dare asking: do you remember the rationale for the fairly cryptic regular expressions ` _DISTRIBUTION_NAME_R` and `_VERSION_R` ?

cournape:
@JCorson sounds good

PR #269, opened 2019-07-12
Remove broken _unsatisfied_clauses cache

### LINKED ISSUE #268: UndeterminedClausePolicy not suggesting best packages
The `UndeterminedClausePolicy.get_next_package_id` method is selecting from a subset of the packages that it should be selecting from.

The apparent cause is that the `_unsatisfied_clauses` attribute is never initialised properly for a set of new packages.

mdickinson:
Fixes #268, by removing the machinery for the `_unsatisfied_clauses` cache. That machinery is broken in two ways:

- the `_unsatisfied_clauses` set isn't ever initialised properly
- the collection of unsatisfied clauses isn't updated when the set of clauses changes (as can happen with the current solver algorithm)

I'd really like to get rid of the changelog too, but we're currently using it for logging purposes.

### DIFF
--- a/simplesat/sat/policy/undetermined_clause_policy.py
+++ b/simplesat/sat/policy/undetermined_clause_policy.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-from collections import defaultdict
-
 import six
 
 from .policy import IPolicy, pkg_id_to_version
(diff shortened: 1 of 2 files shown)


mdickinson:
One other thing: if we *do* want to reimplement something like this cache, I think it would be the solver's job to maintain that cache. We might consider different data structures to organise the information, that would make retrieving the set of unsatisfied clauses more efficient.

mdickinson:
@jvkersch I don't know if you have any bandwidth for review, but if you do I'd very much appreciate a second pair of eyes on this.

Issue #270, opened 2019-10-01
New release of attrs package (19.2.0) breaks simplesat.


benreynwar:
The attrs package recently released a new version (19.2.0) which breaks simplesat.

```
Python 3.7.3 (default, Aug 20 2019, 17:04:43) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import simplesat
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ben/.virtualenvs/deleteme/lib/python3.7/site-packages/simplesat/__init__.py", line 3, in <module>
    from .constraints import Requirement, InstallRequirement
  File "/home/ben/.virtualenvs/deleteme/lib/python3.7/site-packages/simplesat/constraints/__init__.py", line 5, in <module>
    from .constraint_modifiers import (
  File "/home/ben/.virtualenvs/deleteme/lib/python3.7/site-packages/simplesat/constraints/constraint_modifiers.py", line 85, in <module>
    class ConstraintModifiers(object):
  File "/home/ben/.virtualenvs/deleteme/lib/python3.7/site-packages/simplesat/constraints/constraint_modifiers.py", line 86, in ConstraintModifiers
    allow_newer = attr(**_coerced_set)
TypeError: attrib() got an unexpected keyword argument 'convert'
```

itziakos:
fixed by #271

xmentos:
Great and thanks for fixing. Can this be released because it breaks several other packages depending on simplesat (in my case https://github.com/olofk/fusesoc)?

itziakos:
@xmentos  the updated simplesat version 0.8.2 has been uploaded to pypi

PR #271, opened 2019-10-02
MAINT: Fix deprecated attribute 'convert' to comply with attrs 19.2.0

msfschaffner:
The attrs package 19.2.0, which has been released yesterday, deprecates the attribute `convert`. It's basically just renamed to `converter`, so the fix is lightweight. See also

* https://www.attrs.org/en/stable/changelog.html
* https://github.com/python-attrs/attrs/issues/307
* https://github.com/python-attrs/attrs/issues/504

Let me know if I should change anything in this PR.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,10 @@
 Changes since 0.7.0
 ===================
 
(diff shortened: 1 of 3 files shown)


msfschaffner:
This addresses issue #270 .

PR #281, opened 2022-11-18
Replace Travis CI with GitHub Actions

mdickinson:
This PR adds a simple GitHub Actions workflow to run tests and build documentation on each PR. It removes the existing Travis CI configuration.


mdickinson:
@itziakos Would you be interested in reviewing? I think the new GitHub Actions workflow does everything that the old Travis CI config did, albeit with slightly different Python and PyPy versions.

itziakos (approved):

PR #294, opened 2025-03-06
Use raw strings to make Python 3.12 happy

acb-rivos:
Python 3.12 [upgraded a `DeprecationWarning` to a `SyntaxWarning`](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes). These `SyntaxWarnings` appear on stdout for any consuming code.

sat-solver had two `re.compile` calls using normal strings with invalid escape sequences. This PR simply makes those normal strings into raw strings to remedy the issue.

### Testing
Tested locally with Rivos code.


itziakos:
Thanks for the report. I have include/updated the changes in #296.

PR #297, opened 2025-04-02
Better error on cyclic dependencies

itziakos:
- Add support to pass a function to convert package_ids to user friendly strings in toposort.
- Pass a package_id to string function to toposort in the Transaction instance.
- Update tests


dpinte (approved):
LGTM

PR #301, opened 2025-04-02
Fix pypi workflow

itziakos:
- Fix the the pypi workflow
- Delete the download_asset custom action
- Rework the full release workflow
- Minor cleanup

The release ci setup is now as follows:

- On a new release tag the release.yml workflow is triggered and a new draft release is generated after the simplesat solver tests pass
- On a new published release the pipy.yml workflow is triggered and the build-wheel job will build the eggs and upload the archives as artifacts and as release assets on the release tag. Then the pypi-publish job runs to download the artifacts and upload the wheel and sdist to pypi with attastation for github

Note that:
- The wheels are builded and uploaded in a single workflow. 
- The publish job does not have access to the source code.
- Only the publish job has the id-token permissions


itziakos (commented):
PR looks ok

dpinte (approved):
👍

